PR #7, opened 2015-10-31
SDC-1904 - Record EL valueOrDefault should take all types

studanshu:


common/src/main/java/com/streamsets/pipeline/lib/el/RecordEL.java
```diff
@@ -85,7 +85,7 @@ public static Object getValue(
     description = "Returns the value of the field represented by path 'fieldPath' for the record in context or "
       + "the default value if the field is not present")
   public static Object getValueOrDefault(
-    @ElParam("fieldPath") String fieldPath, @ElParam("defaultValue") String defaultValue) {
+    @ElParam("fieldPath") Object fieldPath, @ElParam("defaultValue") Object defaultValue) {
```
brockn: Field path should stay a string as it's always "/something/something"

container/src/test/java/com/streamsets/datacollector/el/TestRecordEL.java
```diff
@@ -69,6 +70,7 @@ public void testRecordFunctions() throws Exception {
     Assert.assertEquals("bar", eval.eval(variables, "${record:attribute('foo')}", String.class));
     Assert.assertEquals("dummy", eval.eval(variables, "${record:valueOrDefault('/z', 'dummy')}", Object.class));
     Assert.assertEquals("A", eval.eval(variables, "${record:valueOrDefault('/a', 'dummy')}", Object.class));
+    Assert.assertEquals("1", eval.eval(variables, "${record:valueOrDefault('/b', 'dummy')}", Object.class));
```
brockn: We want to return the default value here and it shouldn't be a string

container/src/test/java/com/streamsets/datacollector/el/TestRecordEL.java
```diff
@@ -55,6 +55,7 @@ public void testRecordFunctions() throws Exception {
     Mockito.when(record.has(Mockito.eq("/x"))).thenReturn(true);
     Mockito.when(record.has(Mockito.eq("/y"))).thenReturn(false);
     Mockito.when(record.get(Mockito.eq("/a"))).thenReturn(Field.create("A"));
+    Mockito.when(record.has(Mockito.eq("/b"))).thenReturn(Field.create(1));
```
kunickiaj: `Record.has()` has a boolean return type and as such it cannot return a `Field` with a an `int` value of `1` which is what it looks like you're doing here.

Since you're interested in testing that when a field is missing, in this case `record.has("/b") == false` you should return `false` like on line `56`.

PR #12, opened 2016-04-10
SDC-2669: Buffer Limit error interpreted incorrectly

jarcec:


kunickiaj:
This definitely fixes the error message, but there is still the underlying problem that it's hardcoded at 1MB, thus validation fails even if the system limit is set higher than 1MB.

Yet, if you start the pipeline it still starts and processes data > 1MB so I'm trying to understand how that works.

PR #14, opened 2016-07-14
SDC-3435

aharonha:
https://issues.streamsets.com/browse/SDC-3435

### DIFF
--- a/container/src/main/java/com/streamsets/datacollector/record/EventRecordImpl.java
+++ b/container/src/main/java/com/streamsets/datacollector/record/EventRecordImpl.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright 2016 StreamSets Inc.
+ *
+ * Licensed under the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.streamsets.datacollector.record;
+
+import com.streamsets.pipeline.api.EventRecord;
+
+import java.util.Date;
+
+public class EventRecordImpl extends RecordImpl implements EventRecord {
+
+  public EventRecordImpl(String type, int version, String stageCreator, String recordSourceId, byte[] raw, String rawMime) {
+    super(stageCreator, recordSourceId, raw, rawMime);
+    setEventAtributes(type, version);
+  }
+
+  private void setEventAtributes(String type, int version) {
+    getHeader().setAttribute(EventRecord.TYPE, type);
+    getHeader().setAttribute(EventRecord.VERSION, String.valueOf(version));
+    getHeader().setAttribute(EventRecord.CREATION_TIMESTAMP, String.valueOf(System.currentTimeMillis()));
+  }
+
+}
--- a/container/src/main/java/com/streamsets/datacollector/runner/StageContext.java
+++ b/container/src/main/java/com/streamsets/datacollector/runner/StageContext.java
@@ -33,6 +33,7 @@ import com.streamsets.datacollector.el.ELVariables;
 import com.streamsets.datacollector.email.EmailException;
 import com.streamsets.datacollector.email.EmailSender;
 import com.streamsets.datacollector.metrics.MetricsConfigurator;
+import com.streamsets.datacollector.record.EventRecordImpl;
 import com.streamsets.datacollector.record.RecordImpl;
 import com.streamsets.datacollector.record.io.RecordWriterReaderFactory;
 import com.streamsets.datacollector.util.ContainerError;
@@ -380,8 +381,7 @@ public class StageContext implements Source.Context, Target.Context, Processor.C
 
   @Override
   public EventRecord createEventRecord(String type, int version) {
-    // Stub to be implemented as part of SDC-3421
-    return null;
+    return new EventRecordImpl(type, version, instanceName, Utils.format("event:{}:{}:{}:{}", type, version, instanceName, System.currentTimeMillis()), null, null);
   }
 
   @Override
--- a/container/src/test/java/com/streamsets/datacollector/runner/TestStageContext.java
+++ b/container/src/test/java/com/streamsets/datacollector/runner/TestStageContext.java
@@ -32,6 +32,7 @@ import com.streamsets.datacollector.runner.StageContext;
 import com.streamsets.datacollector.util.Configuration;
 import com.streamsets.datacollector.util.ContainerError;
 import com.streamsets.pipeline.api.ErrorCode;
+import com.streamsets.pipeline.api.EventRecord;
 import com.streamsets.pipeline.api.ExecutionMode;
 import com.streamsets.pipeline.api.OnRecordError;
 import com.streamsets.pipeline.api.Record;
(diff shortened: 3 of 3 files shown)


cloudera-integration/pom.xml
```diff
@@ -31,6 +31,12 @@
   <name>StreamSets Parcel and CSD for Cloudera Manager</name>
   <packaging>pom</packaging>
 
+  <repositories>
+    <repository>
+      <id>cloudera</id>
```
jarcec: Out of my curiosity why is repository declaration needed?

But nevertheless, it's not relevant to this patch and hence should be done in separate JIRA/commit.

hdfs-protolib/src/test/java/com/streamsets/pipeline/stage/destination/hdfs/writer/RecordWriterManagerTestBuilder.java
```diff
@@ -40,6 +40,7 @@
   private URI hdfsUri = new URI("file:///");
   private Configuration hdfsConf = new HdfsConfiguration();
   private String uniquePrefix = "prefix";
+  private String fileNameSuffix="";
```
jarcec: We should add method to the builder that will enable changing the suffix.

hdfs-protolib/src/main/java/com/streamsets/pipeline/stage/destination/hdfs/HdfsTargetConfigBean.java
```diff
@@ -139,7 +139,19 @@
...
+  
+  @ConfigDef(
+    required = false,
+    type = ConfigDef.Type.STRING,
+    defaultValue = "",
+    label = "Files Suffix",
+    description = "File name suffix",
+    displayPosition = 106,
+    group = "OUTPUT_FILES",
+    elDefs = SdcEL.class
```
jarcec: I'm wondering if ELs are required for a suffix? Since the file format is a constant, so should be the suffix, right?

hdfs-protolib/src/main/java/com/streamsets/pipeline/stage/destination/hdfs/HdfsTargetConfigBean.java
```diff
@@ -139,7 +139,19 @@
...
   )
   public String uniquePrefix;
-
+  
+  @ConfigDef(
+    required = false,
+    type = ConfigDef.Type.STRING,
+    defaultValue = "",
+    label = "Files Suffix",
+    description = "File name suffix",
```
jarcec: Can we put reasonable example to the description? Like ".txt" so that people knows whether dot is expected or will be added automatically and such.

aws-lib/src/test/java/com/streamsets/pipeline/stage/destination/s3/TestAmazonS3Target.java
```diff
@@ -118,7 +120,7 @@ public void testWriteTextData() throws Exception {
     objectListing = s3client.listObjects(BUCKET_NAME, prefix);
     Assert.assertEquals(1, objectListing.getObjectSummaries().size());
     S3ObjectSummary objectSummary = objectListing.getObjectSummaries().get(0);
-
+    Assert.assertThat("key suffix" , objectSummary.getKey(), CoreMatchers.endsWith(".suffix"));
```
jarcec: Let's also add test for HDFS target.

cloudera-integration/pom.xml
aharonha: The code would not compile without it. Probably you have it on your own repository, making it accessible without the repository declaration. I agree that it is not relevant to this JIRA.

hdfs-protolib/src/main/java/com/streamsets/pipeline/stage/destination/hdfs/HdfsTargetConfigBean.java
aharonha: Copy-Paste typo :-(

hdfs-protolib/src/main/java/com/streamsets/pipeline/stage/destination/hdfs/HdfsTargetConfigBean.java
aharonha: I didn't want to change others activity so I believe that default value should be the empty string.

hdfs-protolib/src/main/java/com/streamsets/pipeline/stage/destination/hdfs/HdfsTargetConfigBean.java
```diff
@@ -526,7 +541,8 @@ public void init(Stage.Context context, List<Stage.ConfigIssue> issues) {
                   new URI(hdfsUri),
                   hdfsConfiguration,
                   uniquePrefix,
-                  false, // Late records doesn't support "template directory" to be in header
```
aharonha: Auto refactoring moved the comment to the other line.

hdfs-protolib/src/test/java/com/streamsets/pipeline/stage/destination/hdfs/writer/RecordWriterManagerTestBuilder.java
```diff
@@ -155,6 +156,7 @@ public RecordWriterManager build() {
       hdfsUri,
       hdfsConf,
       uniquePrefix,
+      fileNameSuffix,
```
aharonha: > We should add method to the builder that will enable changing the suffix. 

@jarcec I think that this line does

cloudera-integration/pom.xml
kunickiaj: Yep, this would be good in a separate JIRA. And you're right we do mirror a bunch of third party repos including this one which is why we didn't catch it.

kunickiaj:
I would also ask that you squash these commits into one, preferably with the first line of the commit msg being <JIRA ID>. <JIRA SUMMARY>

metadaddy:
@kunickiaj "the first line of the commit msg being ." - was there something missing there, or did you literally mean a period?

aws-lib/src/main/java/com/streamsets/pipeline/stage/destination/s3/AmazonS3Target.java
```diff
@@ -66,6 +67,7 @@
   public AmazonS3Target(S3TargetConfigBean s3TargetConfigBean) {
     this.s3TargetConfigBean = s3TargetConfigBean;
     this.partitionTemplate = s3TargetConfigBean.partitionTemplate == null ? "" : s3TargetConfigBean.partitionTemplate;
+    this.fileNameSuffix = s3TargetConfigBean.fileNameSuffix == null ? "" : s3TargetConfigBean.fileNameSuffix;
```
kunickiaj: you can eliminate this null check if you simply initialize filenameSuffix in S3TargetConfigBean.java line 86

aws-lib/src/main/java/com/streamsets/pipeline/stage/destination/s3/AmazonS3Target.java
```diff
@@ -164,7 +166,7 @@ public void write(Batch batch) throws StageException {
         if (writtenRecordCount > 0) {
           fileCount++;
           StringBuilder fileName = new StringBuilder();
```
kunickiaj: nit: you can initialize StringBuilder with the first part of the string to avoid an additional reallocation (which also allocates an extra 16 chars).

`new StringBuilder(keyPrefix)`

It may be even better to simply allocate a "reasonable" amount for the entire filename with all suffixes to avoid resizes (as it'll end up happening frequently if the total length > prefix + 16)

e.g. `new StringBuilder(200)` or some reasonable sounding value.

hdfs-protolib/src/main/java/com/streamsets/pipeline/stage/destination/hdfs/HdfsTargetConfigBean.java
```diff
@@ -434,6 +446,9 @@ public void init(Stage.Context context, List<Stage.ConfigIssue> issues) {
       uniquePrefix = "";
     }
 
+    if (fileNameSuffix == null) {
```
kunickiaj: this block can be removed if you initialize fileNameSuffix in S3TargetConfigBean line 86

PR #19, opened 2016-10-14
MySql Binary Log origin

WgStreamsets:
# MySql BinLog origin

Origin acting as MySql replication slave. It consumes events directly from MySql binary log using library [https://github.com/shyiko/mysql-binlog-connector-java](mysql-binlog-connector).
Events consumed from log are further enriched with columns names and converted to `Record` with appropriate data types.

Origin works well with MySql servers with and without GTID enabled.
## Implementation details and caveats
- GTID support
  Origin works both with GTID enabled and disabled. But if you change GTID mode on/off on the server - please reset origin, as it will not be able to parse offset correctly.
- Origin connects to mysql with `ssl=off`, if you need to change this - please change the source.
- Origin may produce more records than `maxBatchSize`, because individual `update/delete` events may affect many rows and they all are processed as one event. But mysql internally breaks statements affecting many rows to multiple events with moderate rows count in each.
- Origin can work with `binlog_row_image=MINIMAL`, but in this case it does not include default columns values for inserts.
- Origin does not check server's `serverId`, so its up to you to setup offset correctly when connecting to different server if GTID is off. When GTID is on - it does not matter.
- In case of server errors origin stops.
- If mysql connector cannot parse some event (it is known that this happens with `POINT` columns) - error is logged, but origin continues working.
- Knows issue - if you start from beginning on first start and there very schema changes between first binlog event and current time - result may be unpredictable. The same issue happens if you stop origin, make changes to schema (drop a column for example) and then start origin. If it meets some event for table with changed schema - it may produce complete nonsense.


kunickiaj:
Thanks for the contribution! Just wanted to acknowledge that we are reviewing this patch.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
```diff
@@ -0,0 +1,356 @@
...
+                    if (isGtidEnabled()) {
+                        // when starting from beginning with GTID - skip GTIDs that have been removed from server logs already
+                        GtidSet purged = new GtidSet(Util.getServerGtidPurged(dataSource));
+                        // client's gtidset includes first event of purged, skip latest tx of purged
+                        for (GtidSet.UUIDSet uuidSet : purged.getUUIDSets()) {
+                            GtidSet.Interval last = null;
+                            for (GtidSet.Interval interval : uuidSet.getIntervals()) {
+                                last = interval;
+                            }
+                            purged.add(uuidSet.getUUID().toString() + ":" + String.valueOf(last.getEnd()));
```
kunickiaj: getUUID() already returns a string, no need to call toString().

You also dont need String.valueOf() as long will automatically be converted.

(Possibly) need a null check for last, because last.getEnd() could cause NPE if it wasn't initialized in UUIDSet creation

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
```diff
@@ -0,0 +1,356 @@
...
+                        for (GtidSet.UUIDSet uuidSet : ex.getUUIDSets()) {
+                            if (uuidSet.getUUID().equals(serverUUID)) {
+                                List<GtidSet.Interval> intervals = new ArrayList<>(uuidSet.getIntervals());
+                                GtidSet.Interval last = intervals.get(intervals.size() - 1);
+                                ex.add(String.format("%s:%d", serverUUID, last.getEnd()));
+                                break;
+                            }
+                        }
+                        client.setGtidSet(ex.toString());
+                    } else {
```
kunickiaj: dont need an empty else.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
```diff
@@ -0,0 +1,356 @@
...
+        try {
+            return "ON".equals(Util.getGlobalVariable(dataSource, "gtid_mode"));
+        } catch (SQLException e) {
+            throw Throwables.propagate(e);
+        }
+    }
+
+    private void handleErrors() throws StageException {
+        // handle server errors
+        for (ServerException e : serverErrors) {
```
kunickiaj: why the loop? Once you throw the first exception you're going to stop looping, so its not possible to throw more than the first exception from this method.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
```diff
@@ -0,0 +1,178 @@
...
+            group = "CREDENTIALS"
+    )
+    public String password;
+
+    @ConfigDef(
+            required = true,
+            type = ConfigDef.Type.NUMBER,
+            defaultValue = "999",
+            label = "ServerId",
+            description = "ServerId used by binlog client. Must be unique among all replication slaves (origin acts as a replication slave itself).",
```
kunickiaj: should wrap all the long descriptions at 120 chars max

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/SourceOffsetFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package com.streamsets.pipeline.stage.origin.mysql;
+
+public interface SourceOffsetFactory {
+    SourceOffsetFactory GTID = new GtidSourceOffsetFactory();
+    SourceOffsetFactory BIN_LOG = new BinLogPositionOffsetFactory();
+
+    SourceOffset create(String offset);
+
+    class GtidSourceOffsetFactory implements SourceOffsetFactory {
```
kunickiaj: nit: move the implementation to a separate file

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/Util.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package com.streamsets.pipeline.stage.origin.mysql;
+
+import javax.sql.DataSource;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class Util {
```
kunickiaj: add private constructor for utility classes

`private Util() {}`

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/filters/IgnoreTableFilter.java
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+package com.streamsets.pipeline.stage.origin.mysql.filters;
+
+import com.streamsets.pipeline.stage.origin.mysql.EnrichedEvent;
+
+import java.util.regex.Pattern;
+
+/**
+ * Ignores events for given table. Database and table name are case-insensetive.
```
kunickiaj: typo: insensitive

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/filters/IncludeTableFilter.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+package com.streamsets.pipeline.stage.origin.mysql.filters;
+
+import com.streamsets.pipeline.stage.origin.mysql.EnrichedEvent;
+
+import java.util.regex.Pattern;
+
+/**
+ * Accepts events for given table. If database and table name do not match - rejects event.
+ * Database and table name are case-insensetive.
```
kunickiaj: typo: insensitive

mysql-binlog-lib/src/test/java/com/streamsets/pipeline/stage/origin/mysql/GtidSourceOffsetTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Test;
+
+import java.util.UUID;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class GtidSourceOffsetTest {
+    private String gtidSet1 = UUID.randomUUID() + ":1-10";
+    private String gtidSet2 = UUID.randomUUID() + ":2-1100";
```
kunickiaj: i dont see gtidSet2 ever being used.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/BinaryLogConsumer.java
```diff
@@ -0,0 +1,203 @@
...
+        Table table = null;
+        if (tableOpt.isPresent()) {
+            table = tableOpt.get();
+        } else {
+            LOG.error(Errors.MYSQL_002.getMessage(), tableName.getDatabase(), tableName.getTable());
+            // fallback to table without columns names
+            table = new TableWithoutColumnsNames(tableName.getDatabase(), tableName.getTable());
+        }
+        EnrichedEvent enrichedEvent = new EnrichedEvent(event, table, currentOffset);
+        eventBuffer.put(enrichedEvent);
```
WgStreamsets: Good catch, didn't check return result. 
Maybe you could suggest what is the common way to handly this in streamsets? When may the thread be interrupted and what happens afterwards, may the pipeline continue working?

WgStreamsets:
All mentioned issues are fixed, except I didn't change tab ident from 4 to 2 spaces, because it may make your life harder while reviewing the changes in this commit.
I can reformat everything to 2 spaces in the end, when all other problems are worked out.

kunickiaj:
Excellent, thanks!

Feel free to update the spacing to 2 spaces (4 for line continuations). I also found that these 3 files need an Apache license header added as the rat plugin aborted the build.

```
mysql-binlog-lib/src/test/resources/schema.sql
mysql-binlog-lib/src/test/resources/mysql_gtid_on/my.cnf
mysql-binlog-lib/src/test/resources/mysql_gtid_off/my.cnf
```

With that last update I'll run it through our build process to verify.

kunickiaj:
Sorry about the delay getting back to you on this once again.

I did notice that when running the tests I do get two failures. Could you verify that this also happens in your environment and address?

```
Failed tests:
  MysqlGtidOffSourceIT>AbstractMysqlSource.shouldCreateRecordWithoutColumnNamesWhenMetadataNotFound:501
Expected: is <Field[STRING:2016-08-18]>
     but: was <Field[STRING:2016-08-17]>
  MysqlGtidOnSourceIT>AbstractMysqlSource.shouldCreateRecordWithoutColumnNamesWhenMetadataNotFound:501
Expected: is <Field[STRING:2016-08-18]>
     but: was <Field[STRING:2016-08-17]>

Tests run: 31, Failures: 2, Errors: 0, Skipped: 0
```

kunickiaj:
Interestingly enough, still getting off-by-one:

```
Failed tests:
  MysqlGtidOffSourceIT>AbstractMysqlSource.shouldCreateRecordWithoutColumnNamesWhenMetadataNotFound:531
Expected: is <Field[STRING:05:01:02]>
     but: was <Field[STRING:04:01:02]>
  MysqlGtidOnSourceIT>AbstractMysqlSource.shouldCreateRecordWithoutColumnNamesWhenMetadataNotFound:531
Expected: is <Field[STRING:05:01:02]>
     but: was <Field[STRING:04:01:02]>

Tests run: 31, Failures: 2, Errors: 0, Skipped: 0
```

WgStreamsets:
There is some mess with DST in MySql for time fields. I removed the failing check - it is far from important and useful.

mysql-binlog-lib/README.md
```diff
@@ -0,0 +1,32 @@
...
+- Origin may produce more records than `maxBatchSize`, because individual `update/delete` events may affect many rows and they all are processed as one event. But mysql internally breaks statements affecting many rows to multiple events with moderate rows count in each.
+- Origin can work with `binlog_row_image=MINIMAL`, but in this case it does not include default columns values for inserts.
+- Origin does not check server's `serverId`, so its up to you to setup offset correctly when connecting to different server if GTID is off. When GTID is on - it does not matter.
+- In case of server errors origin stops.
+- If mysql connector cannot parse some event (it is known that this happens with `POINT` columns) - error is logged, but origin continues working.
+- Knows issue - if you start from beginning on first start and there very schema changes between first binlog event and current time - result may be unpredictable. The same issue happens if you stop origin, make changes to schema (drop a column for example) and then start origin. If it meets some event for table with changed schema - it may produce complete nonsense.
```
harishreedharan: Known

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/BinaryLogConsumer.java
```diff
@@ -0,0 +1,214 @@
...
+        return;
+      } else {
+        // record current gtid + seqNo as incomplete tx
+        currentOffset = ((GtidSourceOffset) currentOffset)
+            .withIncompleteTransaction(currentTxGtid, currentTxEventSeqNo)
+            .withGtidSet(currentGtidSet);
+      }
+    } else {
+      // current offset is null or gtid mode off
+      currentOffset = createOffset(event);
```
harishreedharan: So this is something I am worried about. If GTID is not enabled, does it mean that a restart in the middle of a transaction might lose some changes from the transaction?

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/BinLogPositionSourceOffset.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+package com.streamsets.pipeline.stage.origin.mysql;
+
+import com.github.shyiko.mysql.binlog.BinaryLogClient;
+
+public class BinLogPositionSourceOffset implements SourceOffset {
+  private final long position;
+  private final String filename;
+
+  public BinLogPositionSourceOffset(String filename, long position) {
```
WgStreamsets: no, they cannot be null. Removed checks in hashCode/equals

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/BinaryLogConsumer.java
harishreedharan: So in that case, we could end up reading uncommitted transactions? Or only committed transactions actually make it to the binary log (I would assume this is not the case)?

WgStreamsets: In my understanding only committed transactions are put to replication binlog. I've never seen a rolled back or partial transaction in replication binlog with InnoDB engine. I did some manual tests with long transactions affecting lots of data, but they always are put to binlog only after commit. 
Though I'm not  a mysql expert and may be missing some cases.

WgStreamsets: but problems may happen when mixing transactional and nontransactional statements in one tx - http://dev.mysql.com/doc/refman/5.7/en/replication-features-transactions.html

harishreedharan: That should be fine no? If a non-transactional statement gets rolled back manually, that would appear in a separate binlog entry anyway. So we should be ok.

If that is the case, this looks good to me.

harishreedharan:
+2.

@kunickiaj This looks good to me. If all changes you requested are in, we can go ahead and merge this.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/BinaryLogConsumer.java
WgStreamsets: Yes, for us it is okay.

kunickiaj:
merged.

PR #20, opened 2016-11-10
Convert Long to DateTime before String

SandishKumarHN:
Checkbox Selection to convert input Long to DateTime before converting to a String.

User Story:

My data has an incoming long value which represents a timestamp. I want the StreamSets "Type Converter" to format that value as a string. For example, I'd to be able to take a field which is of long time, valued 1478298364 and have the type converter convert the field to "2016-11-04."

Design:

The field converter already allows us to specify a string and a date format. The problem is that when you have a long filed you must first convert to datetime then to string, which may not be obvious to the user:

[https://raw.githubusercontent.com/SandishKumarHN/Hadoop/master/Streamsets_LongtoDate.png](url)

As such, we should add a checkpoint that is only shown when the the target is of type String which says "Treat incoming long type as DateTime" and when that checkbox is checked, we convert the long field to datetime then to a string.


rcprcp:
hi SandishKumarHN, 

Thank you for your contribution.  I quickly reviewed your contribution, and it looks good.  

Please ensure you have filled out and submitted the CLA or CCLA (there are separate agreements for individuals or corporations) - additional information can be found here: 
[https://streamsets.com/contributing/](url)

So far, i think there are a couple of minor issues, mostly style related.  Please feel free to review the style guide: [https://github.com/streamsets/datacollector/wiki/Code-Style-Guide](url)   The main issues are that we use 2 spaces as “tabs”, and the max line width should be capped at 120.  

The changes in BaseConverterConfig.java are indented too much.  Also, the {} around triggeredByValue={“STRING”} are not required.  

Your new section in FieldTypeConverterProcessor.java has a couple of very wide lines (the if statement and the LOG statement).  I’ve also noticed that there are a bunch of wide lines in the file already, at some point we’ll reformat it to be compliant.   :)  

Please review and “sign” the CLA/CCLA, and then we will have a more in depth look at the patch for additional review and testing.

Thanks again for your contribution, it is greatly appreciated!!

bob

SandishKumarHN:
Thanks for the review, just signed CLA/CCLA from here https://streamsets.com/contributing/, made style changes too, please check

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -47,7 +47,19 @@
   )
   @ValueChooserModel(PrimitiveFieldTypeChooserValues.class)
   public Field.Type targetType;
-
+  
+  @ConfigDef(
+	      required = true,
+	      type = ConfigDef.Type.BOOLEAN,
+	      defaultValue = "false",
+	      label = "Convert Long to DateTime before String",
```
kunickiaj: How about something more along the lines of "Treat Input Field as Date".
This then acts as a hint that we should take the input and treat it as a date.
For example if its a long we convert to DateTime then apply the string date formatter.
If its already a DateTime/Date field we apply the string date formatter.
If its something we don't know how to convert to a Date, then send the record to the error pipeline.

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -47,7 +47,19 @@
...
+	      description = "Select to convert input Long to DateTime before converting to a String",
+	      displayPosition = 20,
+	      dependsOn = "targetType",
```
kunickiaj: Right it it seems like we always display the Date Formatter options when String is selected for target field type.

Since now we support multiple dependsOn relationships, it would be nice (and probably less confusing for users) to also hide the Date Format options unless "Treat Input Field as Date" is selected.

An example of using multiple dependencies can be seen here:
https://github.com/streamsets/datacollector/blob/master/commonlib/src/main/java/com/streamsets/pipeline/stage/destination/lib/DataGeneratorFormatConfig.java#L279-L282

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -103,8 +116,10 @@ public Locale getLocale() {
       label = "Date Format",
       description="Select or enter any valid date or datetime format",
       displayPosition = 40,
-      dependsOn = "targetType",
-      triggeredByValue = {"DATE", "DATETIME", "TIME", "STRING"}
+      dependencies = {
+          @Dependency(configName = "targetType", triggeredByValues = {"DATE", "DATETIME", "TIME", "STRING"}),
```
SandishKumarHN: Please correct me on this, can we make dateformat field option without treatInputFieldAsDate checkbox condition for "DATE","DATETIME","TIME".

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -47,7 +48,19 @@
...
+  
+  @ConfigDef(
+	      required = true,
+	      type = ConfigDef.Type.BOOLEAN,
+	      defaultValue = "false",
+	      label = "Treat Input Field as Date",
+	      description = "Select to convert input to DateTime before converting to a String",
+	      displayPosition = 20,
+	      dependsOn = "targetType",
+	      triggeredByValue = {"DATE", "DATETIME", "TIME", "STRING"}
```
SandishKumarHN: Please correct me on this, can we make dateformat field option without treatInputFieldAsDate checkbox condition for "DATE","DATETIME","TIME".

kunickiaj: I'm not sure I follow.

The treatInputFieldAsDate checkbox condition should only apply to target type STRING.

By selecting DATE or DATETIME as the target field, converting from a long already works correctly IIRC, thus there's no need to show the checkbox there.

Assuming the checkbox also is the trigger for showing the DateFormat options, that should work also, because DateFormat only applies to target format String.

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -103,8 +116,10 @@ public Locale getLocale() {
       label = "Date Format",
       description="Select or enter any valid date or datetime format",
       displayPosition = 40,
-      dependsOn = "targetType",
-      triggeredByValue = {"DATE", "DATETIME", "TIME", "STRING"}
+      dependencies = {
```
SandishKumarHN: How can I trigger DATEFOMRAT when targetType is "DATE", "DATETIME", "TIME" without using treatInputFieldAsDate value??
or
How can I trigger DATEFORMAT when targetType is "STRING" and treatInputFieldAsDate is true?

Is there any way??

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -103,8 +116,10 @@ public Locale getLocale() {
...
+          @Dependency(configName = "targetType", triggeredByValues = {"DATE", "DATETIME", "TIME", "STRING"}),
+          @Dependency(configName = "treatInputFieldAsDate", triggeredByValues = "true")
+      }
```
SandishKumarHN: How can I trigger DATEFOMRAT when targetType is "DATE", "DATETIME", "TIME" without using treatInputFieldAsDate value??
or
How can I trigger DATEFORMAT when targetType is "STRING" and treatInputFieldAsDate is true?

Is there anyway??

SandishKumarHN:
I see there are no AND+OR  relationships with dependencies so, I created duplicates of DATEFORMAT and OTHERDATEFORMATE configs

kunickiaj:
@brockn I figured this was a pretty simple change but clearly I'm misunderstanding something here.

I just fired up SDC to take a closer look. It appears that we already do hide the DateFormat except when converting to String or to Date/DateTime (I assume that means its used only when the input is a String).

I see that this is the reason for the otherDateFormat config (no OR). I'm fine if we just want to keep the original scope as this seems to be more complicated than I anticipated.

SandishKumarHN:
Reverted back to original change with label rename to "Treat Input Field As Date"

basic-lib/src/test/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/TestFieldTypeConverterProcessorFields.java
```diff
@@ -971,7 +971,52 @@ public void testStringToDate() throws Exception {
...
 
+    ProcessorRunner runner = new ProcessorRunner.Builder(FieldTypeConverterDProcessor.class)
+        .addConfiguration("convertBy", ConvertBy.BY_FIELD)
+        .addConfiguration("fieldTypeConverterConfigs", ImmutableList.of(stringConfig, dtConfig))
+        .addOutputLane("a").build();
+    runner.runInit();
+
+    try {
+      Map<String, Field> map = new LinkedHashMap<>();
+      map.put("longString", Field.create(1431763314761l));
```
kunickiaj: nit: please use `L` instead of `l` since with some fonts it looks a lot like a `1`

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -47,7 +47,19 @@
   )
   @ValueChooserModel(PrimitiveFieldTypeChooserValues.class)
   public Field.Type targetType;
-
+  
+  @ConfigDef(
+	      required = true,
```
kunickiaj: these should be spaces, not tabs.

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/BaseConverterConfig.java
```diff
@@ -47,7 +47,19 @@
...
+	      type = ConfigDef.Type.BOOLEAN,
+	      defaultValue = "false",
+	      label = "Treat Input Field as Date",
+	      description = "Select to convert input Long to DateTime before converting to a String",
+	      displayPosition = 20,
+	      dependsOn = "targetType",
+	      triggeredByValue = "STRING"
+	  )
```
kunickiaj: closing paren should match indent level of opening paren. see https://github.com/streamsets/datacollector/wiki/Code-Style-Guide for example

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/FieldTypeConverterProcessor.java
```diff
@@ -139,6 +139,17 @@ private Field convertField(String matchingField, Field field, BaseConverterConfi
       }
     }
 
+    if(converterConfig.targetType.isOneOf(Field.Type.STRING) && converterConfig.treatInputFieldAsDate && field.getType() == Field.Type.LONG) {
```
kunickiaj: please wrap lines > 120 chars.

SandishKumarHN:
Done!

kunickiaj:
LGTM

kunickiaj:
Hey, sorry, just realized that this change also requires a StageUpgrader as it adds new configuration values. This is necessary to migrate pipelines previously created to also add this stage configuration.

There are plenty of upgrader examples in the codebase such as https://github.com/streamsets/datacollector/blob/master/jython-protolib/src/main/java/com/streamsets/pipeline/stage/processor/jython/JythonProcessorUpgrader.java

SandishKumarHN:
Ok will do

SandishKumarHN:
Please let me know, If my changes are incorrect.

basic-lib/src/main/java/com/streamsets/pipeline/stage/processor/fieldtypeconverter/FieldTypeConverterProcessorUpgrader.java
```diff
@@ -0,0 +1,38 @@
+package com.streamsets.pipeline.stage.processor.fieldtypeconverter;
```
brockn: Needs the copyright and license

SandishKumarHN: Thanks Added.

brockn: Copyright year is old :)

SandishKumarHN: Sorry, changed to 2016 year

brockn:
@kunickiaj - this should be good to go?

kunickiaj:
Hey, yeah this looks good to go. I'll put it through the build process. Sorry for the delay, was out of pocket for 10 days.

kunickiaj:
Merged

PR #23, opened 2017-02-16
Added the option to parse an array of json objects to the data format…

lordjc:
… config of httpclient


jarcec:
I wonder if you have seen my feedback @lordjc ?

lordjc:
I did see your feedback but will need to take a closer look. When I tested
I still hit an issue so I am not sure if you are saying that if I rebase it
will fix the issue that I hit?

jarcec:
I'm not sure what "issue" you're hitting @lordjc, all I see is the patch, and that seems to be allowing additional options on the HTTP origin which seem like a fair change.

jarcec:
It seems that this patch is clearly trying to un-hide a property that is generally helpful, so let's perhaps get this in and let's investigate the troubles you're observing separately @lordjc ?

jarcec:
I'll close this one as there hasn't been any update since last August, please don't hesitate and reopen or create a new pull request!

PR #24, opened 2017-03-02
Added null check for Kafka message payload

chadphillips:
When consuming from a compacted Kafka topic, some messages will contain a null payload to indicate that messages with that key will be removed in the future when the log is compacted.  A null check is required to deal with those message payloads gracefully rather than throwing an NPE.


jarcec:
Have you signed our ICLA @chadphillips ?

https://streamsets.com/contributing/

Otherwise looks good to me.

chadphillips:
@jarcec Yes, the ICLA I submitted was singed by Jobi George on 2017-03-13.

kunickiaj:
Committed as SDC-6787. Thanks for the contribution!

PR #32, opened 2017-06-19
Kudu Lookup Processor

brockn:
* All primary keys on the target table must be specified in the record
 or a record error is generated
* Unit tests need to be expanded
* Never tested against a secure cluster

### DIFF
--- a/kudu-protolib/pom.xml
+++ b/kudu-protolib/pom.xml
@@ -76,6 +76,13 @@
     </dependency>
 
     <dependency>
+      <groupId>com.streamsets</groupId>
+      <artifactId>streamsets-datacollector-lookup-protolib</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+
+    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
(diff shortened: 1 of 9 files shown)


brockn:
Hey guys,

Here is the basic functionality. It's admitted a little rough and could use some polishing. I am unfortunately take some vacation for the next two weeks so if someone has time, please feel free to pickup and finish off. Otherwise I can pickup in July.

brockn:
I noticed that "fieldMappingConfigs" still says it's optional. It's not for the processor since we don't want to pull back the entire row, that's how we find the projection.

metadaddy:
Cool - looking forward to those expanded unit tests!

ujunko:
Hi Brock! We have specific use case that we'd like to cover - receiving primary key values from Kudu lookup. This patch requires primary key values in order to perform lookup. I'm wondering if we can add predicates from field columns rather than primary keys (load() in KuduLookupLoader).

brockn:
"This patch requires primary key values in order to perform lookup. I'm wondering if we can add predicates from field columns rather than primary keys (load() in KuduLookupLoader)."

Hmm without the primary keys you have to do a full-table scan so we should probably discuss if this is a good idea. Without primary key the table would have to be very small.

jarcec:
@ujunko have merged this early today, so I'll close the pull request.

PR #35, opened 2017-08-05
Patch for allowing users to supply / extract custom query parameters …

sreeraaman:
We have a pipeline with http server as origin and a custom processor for normalising csv records. The normalisation happens via a rest call that in turn returns the file format definition for the file being normalized.

As such we need to send two parameters one is the filename and the other being the pipeline mode (whether it is a dry run or a live upload)

based on the filename, we retrieve the metadata about the file format, field definitions etc.
based on the pipeline mode we decide whether to push the normalized record into the live topic or dry run topic.

presently, the http server origin does not provide any information about the file being uploaded nor it passes on the query parameters to record headers to other stage components.

This patch addresses this issue.
Attaching the sample pipeline.
pipeline

### DIFF
--- a/basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/httpserver/PushHttpReceiver.java
+++ b/basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/httpserver/PushHttpReceiver.java
@@ -124,6 +124,15 @@ public class PushHttpReceiver implements HttpReceiver {
 
     // parse request into records
     List<Record> records = new ArrayList<>();
+
+    //begin fix for obtaining custom query parameters.
+    Map<String,Object> httpQueryParams = new HashMap<>();
+    req.getParameterMap().entrySet().forEach(entry ->{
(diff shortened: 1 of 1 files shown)


metadaddy:
Hi @sreeraaman - have you signed the StreamSets CLA? https://streamsets.com/contributing/

sreeraaman:
Hi,
I have signed the Streamsets CLA.
Thanks
Sriraman.

jarcec:
Thanks for the contribution @sreeraaman ! Few high-level questions:

* I think that we currently put the HTTP response headers to our headers rather than the URL arguments, so I wonder what happens if there is a collision?
* I believe that the HTTP request params can contain sensitive information (passwords, keys), so I wonder if we should be somehow smart and not leak them?

Are you open to sharing your whole pipeline? Since the HTTP Client Origin must be able to calculate the request header attributes from constants (it's an origin, so RecordEL doesn't make sense there), I wonder if you could easily use the same logic on your custom Normalizer component.

jarcec:
There is no update since late August, so I'll close the pull request for now. Please don't hesitate and re-open or submit a new one!

PR #38, opened 2017-10-02
connectionTimeout input value is in seconds, while jedis expects milliseconds

lingma:
connectionTimeout input value is in seconds, while jedis expects milliseconds, a multiplication of 1000 is necessary.


lingma:
I have another file to request

PR #40, opened 2017-10-30
SDC-5463. Changed rabbit-mq expiration type from short to int

ArunTejCh:
Int should be fine for all 4 usages of this variable. RabbitMQ's short str has a limit of 255 chars. String.valueOf(expiration) should not cause any problems when expiration has type int.


jarcec:
I wonder if you signed our ICLA already @ArunTejCh ?

https://streamsets.com/contributing/

jarcec:
Do you mind the ICLA @ArunTejCh - I see other contributors working on a similar change and I would like to merge yours as you were first one to fix it. Sadly I do need you to sign the ICLA first as without it I can't merge the change.

jarcec:
I will resolve the pull request for now as we're missing ICLA signature for a long time.

PR #41, opened 2017-11-02
Pull to Windows: fatal: cannot create directory at 'aux': Invalid argument

bosik:
As `aux` is reserved word in Windows, having `aux` folder in repository makes clonning it in Windows impossible (clonning fails with "fatal: cannot create directory at 'aux': Invalid argument.").

As a solution, `aux` folder was renamed to `misc` + script references updated.

Please check if there are any other references / if build is ok, as I'm not able to test it myself.


kunickiaj:
I don't believe this can be helped as aux is required as part of the structure of the CSD. ref: https://github.com/cloudera/cm_ext/wiki/The-Structure-of-a-CSD

Is this for development on Windows? Windows is not currently a supported platform for SDC, though I've had some success running it under Windows Subsystem for Linux (Available on Windows 10 Pro and Windows Server Insider builds)

jarcec:
There is also quite old JIRA with some background [SDC-4809](https://issues.streamsets.com/browse/SDC-4809). As @kunickiaj mentioned, this name is imposed on us by partner integration, so not much we can do about it.

bosik:
@kunickiaj I realize some more changes might be required, as I'm new to CSD and don't see big picture. Nevertheless, I think the change is useful for any Windows developer, because the issue totally blocks clonning, and hence, any development.

@jarcec Thanks for pointing. Seems there is no chance :(

metadaddy:
Thinking out of the box a little here, @kunickiaj / @jarcec, would it be possible to call the directory `aux` on disk, but rename it as it's built into the CSD?

jarcec:
I will go ahead and resolve this pull request for now since we don't have a plan to change the build system and the aux library is indeed required by Cloudera.

PR #45, opened 2017-11-21
work around UnknownHostException

m-khl:
Hello!
Thanks for the great product.

I've found one issue on AWS. Some EC2 instances has odd VPC setup like here https://forums.aws.amazon.com/thread.jspa?threadID=164739 (requires aws auth; copied it below). InetAddress.getLocalHost() always throws UnknownHostException on such instances, I call it "poisoned". It doesn't seem problematic, since I can set hostname in sdc.properties. But the `RuntimeEL.loadConficuration` calls InetAddress.getLocalHost() to obtain default value before reading the configured property. It's sad. If have sdc.properties I even can't run tests on such instances.

The proposed patch let it pull default hostname value lazily, only if it's absent in sdc.properties. So, I can prevent calling InetAddress.getLocalHost() on poisoned instances by supplying a hostname in sdc.properties.  
I have to make a buch of other changes to provide a test for it. 

Thanks for your time! Keep going.

### DIFF
--- a/container/src/main/java/com/streamsets/datacollector/el/RuntimeEL.java
+++ b/container/src/main/java/com/streamsets/datacollector/el/RuntimeEL.java
@@ -15,6 +15,7 @@
  */
 package com.streamsets.datacollector.el;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.streamsets.datacollector.http.WebServerTask;
 import com.streamsets.datacollector.main.RuntimeInfo;
 import com.streamsets.datacollector.util.Configuration;
@@ -40,6 +41,11 @@ import java.util.Set;
 
 public class RuntimeEL {
 
+  @FunctionalInterface
+  interface  Supplier<R,E extends Exception> {
+      R get() throws E;
+  }
+
   private static final Logger LOG = LoggerFactory.getLogger(RuntimeEL.class);
   private static final String SDC_PROPERTIES = "sdc.properties";
   private static final String RUNTIME_CONF_LOCATION_KEY = "runtime.conf.location";
@@ -49,6 +55,9 @@ public class RuntimeEL {
   private static String AUTH_TOKEN = null;
   private static String HOSTNAME = null;
   private static RuntimeInfo runtimeInfo;
+  
+  @VisibleForTesting
+  static Supplier<String, IOException> defaultHostnameMethod = RuntimeEL::getDefautHostName;
 
   @ElConstant(name = "NULL", description = "NULL value")
   public static final Object NULL = null;
@@ -134,7 +143,10 @@ public class RuntimeEL {
       }
 
       AUTH_TOKEN = runtimeInfo.getAppAuthToken();
-      HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST, InetAddress.getLocalHost().getCanonicalHostName());
+      HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST, null);
+      if (HOSTNAME==null) {
+          HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST,defaultHostnameMethod.get());
+      }
 
       RUNTIME_CONF_PROPS = new Properties();
       String runtimeConfLocation = configuration.get(RUNTIME_CONF_LOCATION_KEY, RUNTIME_CONF_LOCATION_DEFAULT);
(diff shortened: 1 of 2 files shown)


m-khl:
copying aws forum thread 

> Hi there,

I have a VPC setup with one public instance. I was under the impression that resolving internal hostname should work. The instance's external hostname does resolve perfectly (from external clients). The VPC's "DNS resolution" and "DNS hostnames" are both enabled.

Am I correct assuming that internal hostname resolution works?

# hostname
ip-172-30-0-139
 
# ping `hostname`
ping: unknown host ip-172-30-0-139
...

This was happening to me too for instances in my non-default VPC. The "DNS Hostnames" setting is turned off by default for non-default VPCs.

Just had to turn on "DNS Hostnames" for my VPC (from the AWS Console) and my auto-assigned `computer.internal` hostname started resolving almost immediately:

container/src/main/java/com/streamsets/datacollector/el/RuntimeEL.java
```diff
@@ -134,7 +139,7 @@ public static void loadRuntimeConfiguration(RuntimeInfo runtimeInfo) throws IOEx
       }
 
       AUTH_TOKEN = runtimeInfo.getAppAuthToken();
-      HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST, InetAddress.getLocalHost().getCanonicalHostName());
+      HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST, defaultHostnameMethod);
```
jarcec: Wouldn't be a bit simpler and a bit less intrusive to do something in the line of:

```
String defaultHostname = "";
try {
defaultHostname = InetAddress.getLocalHost().getCanonicalHostName());
} catch (...) {}
HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST, defaultHostname);
```

ebs-mkhludnev: Sure. I just tried to practice Java 8 features.  Is it worth to amend the patch by this way?

jarcec: I feel that the current approach is adding a lot of code which will be harder to follow - let's just use the simpler solution here if you don't mind. I'm sure that there will be better places to practice the Java 8 features.

jarcec:
Once the patch gets updated, I'll also need you to sign our [contributor agreement](https://streamsets.com/contributing/) before I'll be able to merge it @ebs-mkhludnev .

container/src/main/java/com/streamsets/datacollector/el/RuntimeEL.java
ebs-mkhludnev: ok. @jarcec, can you clarify, you want just keep util/Configuration.java unchanged and still have functional juggling for sake of testing? or you want drop testHostnamePoisioned() as well?

jarcec: I'm fine with having test @ebs-mkhludnev , but I would prefer to keep the Configuration class unchanged.

m-khl:
@jarcec, no changes in Configuration, test for poisoned hostname is here.

jarcec:
I might have confused you @m-khl with my earlier answer. My main concern is that the Java-8 method reference seems to be an overkill for the problem at hand which will make it harder to maintain in the future. I would really like to simplify it to keep the maitenance burder low. 

What about simply changing the 

```
if (HOSTNAME==null) {
  HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST,defaultHostnameMethod.get());
}
```

To something like the following:

```
if (HOSTNAME==null) {
  HOSTNAME = configuration.get(WebServerTask.HTTP_BIND_HOST, null);
  if(HOSTNAME == null) {
    HOSTNAME = InetAddress.getLocalHost().getCanonicalHostName();
  }
}
```

I believe that this way your original issue will be resolved since we will not call InetAddress.getLocalHost() before ensuring that we don't have HTTP_BIND_HOST and at the same time the change is fairly small for this edge case.

What do you think?

m-khl:
Agree. I need it for testing. On testHostnamePoisioned() I can reproduce that UnknHostExp by replacing method reference. What you propose will work fine, but doesn't have test coverage.

jarcec:
For the test you could use PowerMock that afaik we are already depending on to make sure that InetAddress.getLocalHost() will throw an exception and ensure that it won't cause any issues.

m-khl:
aghhh. I see. Ok I'll kick tires with PowerMocks, but it will take some time. I'll get back in a few days.

jarcec:
Did you by any chance had time to follow up @m-khl ?

m-khl:
Nope. Sorry. Seems like I'm off.

On Fri, Jun 8, 2018 at 5:27 PM, Jarek Jarcec Cecho <notifications@github.com

jarcec:
I will go ahead and resolve the pull request for now. Feel free to re-submit if you will get chance to incorporate the feedback @m-khl !

PR #46, opened 2017-11-23
ANS-16. Add option to CDC config TAB to pass unsupported parameter as string instead of throw exception

estatic:
These changes allows to pass unsupported data type as string without throwing exceptions.


harishreedharan:
We already added this functionality in master and SDC 3.0. Does it not work for you?

https://github.com/streamsets/datacollector/blob/master/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java#L683

estatic:
Hi @harishreedharan,
 
yes, you are right, I've made changes based on 2.7 version. Seems, my changes doesn't need any more.

harishreedharan:
Let me know how 3.0 works for you with respect to this feature.

estatic:
@harishreedharan Yes, this works as we've expected. Thanks a lot!

PR #48, opened 2017-12-14
Feature/es merge operation

ml10:
Elasticsearch supports a "document level upsert" which is basically a merge operation:
- If the document does not exist during a merge operation, it is created/inserted as is.
- If the document does exist in the index during a merge operation, then all of the fields in the incoming document are merged with the existing document. Any fields with the same key take the value of the incoming document, any fields in the existing document not in the incoming document are left as is. This behavior works with nested fields as well.

Currently Streamsets does not support this operation in the Elasticsesarch destination.

This PR adds a new top-level `MERGE OperationType`, that is leveraged by the `ElasticsearchTarget` to support document update using the `doc_as_upsert` value in the request.

Our team is currently using 2.7.X line, but if you'd prefer these changes against the current development target, let me know.


metadaddy:
Thanks for the contribution, @ml10! Please confirm that you (or your company) have signed the CLA: https://streamsets.com/contributing/

Once that's in place, we can have an engineer review your changes.

ml10:
I thought we had a corporate CLA in place already, and it sounds like that is something we want to do. Unfortunately with the current holiday schedule we probably won't get to that until the new year. If you'd rather this PR not hang out that long, I can work on a personal CLA, otherwise we can revisit after the holidays.

metadaddy:
Probably best to get the CCLA in place after the holidays; it's unlikely anyone would get to this before Jan anyway.

jarcec:
The change looks good to me, so let me know when the CCLA gets signed and we'll push it through @ml10 !

ml10:
Thank you for your patience! I got word that Spantree has signed a Corporate CLA that should cover this contribution. Let me know if you need anything further from me. Thanks!

jarcec:
Do you mind creating a user account on our [public JIRA](https://issues.streamsets.com) and creating a new JIRA for this one @ml10 ?

ml10:
Closing this PR in favor of PR #52

PR #42, opened 2017-11-15
Fix issues SDC-7902 and SDC-7861. Also, change apache-rat check start principle.

estatic:
I've tried to build it on my Macbook and found that there are some issues with tests. So, I've create bug report on JIRA and fixed found issues by these commits.
bug SDC-7902 and SDC-7861

Also, I've removed ANT script which generates bash script and change it with exec goal. It allows to build and test on non-UNIX OSes.

PR #50, opened 2017-12-22
SDC-7902. Change server creation by giving static localhost server ad…

estatic:
Somehow it sometimes set wrong IP, so then tests fails. I've set localhost as server.
Moved from https://github.com/streamsets/datacollector/pull/42


jarcec:
Looks good to me @estatic - let me get it merged.

jarcec:
The change was merged, thank you for your contribution @estatic !

PR #51, opened 2017-12-22
SDC-8104. Remove generation OS-depended bash script and ANT command. Replace wi…

estatic:
Moved from https://github.com/streamsets/datacollector/pull/42

Using execution through maven avoid from OS depended issues problems. No need to generate additional OS depended script.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
     <stats-lib>stats-lib</stats-lib>
     <vault-credentialstore-lib>vault-credentialstore-lib</vault-credentialstore-lib>
     <windows-lib>windows-lib</windows-lib>
-
+    <rootProject>true</rootProject>
   </properties>
 
   <!-- StreamSets Data Collector API being used -->
@@ -315,6 +315,7 @@
             <exclude>common-ui/src/main/webapp/common/directives/**</exclude>
 
           </excludes>
+          <goal>run</goal>
         </configuration>
       </plugin>
     </plugins>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -84,7 +84,7 @@
     <stats-lib>stats-lib</stats-lib>
     <vault-credentialstore-lib>vault-credentialstore-lib</vault-credentialstore-lib>
     <windows-lib>windows-lib</windows-lib>
-
+    <rootProject>true</rootProject>
```
jarcec: This is a bit higher maven magic that I'm used to - do you mind walking me through why this property is needed here? It seems that the "rootProject" is not used anywhere in this patch.

pom.xml
```diff
@@ -370,38 +371,33 @@
...
+                <id>rat-check</id>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>run</goal>
+                  <goal>exec</goal>
                 </goals>
-                <configuration>
-                  <target>
-                    <echo file="${project.build.directory}/apache-rat.sh">#!/bin/bash
-                      if [ "${project.artifactId}" == "streamsets-datacollector" ];
```
jarcec: The original semantic was to run the rat check only as part of the root execution - I think that the thinking went that our build is quite long and hence it failing because of rat somewhere in the middle would be time-consuming. This way the build fails almost immediately. I wonder, does the patch preserve the semantic?

estatic:
Hi @jarcec,

Answering to both question.

rootProject is need to execute maven goal only on root project, so as I
tested it works like generated bash script. Means starts only for root
project and scan whole sources.

This patch should allow to move development from only Linux to wide range
of operation systems.

Cheers,
Ivan

26 дек. 2017 г. 5:50 PM пользователь "Jarek Jarcec Cecho" <
notifications@github.com> написал:

jarcec:
Wonderful thank you for the explanation @estatic  - let me run the patch by our CI server and merge it.

jarcec:
And this one got just merged as well @estatic ! Thanks for all three contributions today :)

Happy (late) Christmas and Merry New Year!

estatic:
Merry xmas and happy new year to you @jarcec and to your team :)

26 дек. 2017 г. 7:07 PM пользователь "Jarek Jarcec Cecho" <
notifications@github.com> написал:

Closed #51 <https://github.com/streamsets/datacollector/pull/51>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/streamsets/datacollector/pull/51#event-1401724580>, or mute
the thread

PR #52, opened 2018-01-10
Feature/es merge operation 3.x

ml10:
Elasticsearch supports a "document level upsert" which is basically a merge operation:

- If the document does not exist during a merge operation, it is created/inserted as is.
- If the document does exist in the index during a merge operation, then all of the fields in the incoming document are merged with the existing document. Any fields with the same key take the value of the incoming document, any fields in the existing document not in the incoming document are left as is. This behavior works with nested fields as well.

Currently Streamsets does not support this operation in the Elasticsesarch destination.

This PR adds a new top-level `OperationType.MERGE`, that is leveraged by the `ElasticsearchTarget` to support document update using the `doc_as_upsert` value in the request.

This PR is an update of PR #48, but based off of master instead of the 2.7 series.

### DIFF
--- a/commonlib/src/main/java/com/streamsets/pipeline/lib/operation/OperationType.java
+++ b/commonlib/src/main/java/com/streamsets/pipeline/lib/operation/OperationType.java
@@ -30,6 +30,7 @@ public class OperationType {
   public static final int UNSUPPORTED_CODE = 5;
   public static final int UNDELETE_CODE = 6;
   public static final int REPLACE_CODE = 7;
+  public static final int MERGE_CODE = 8;
 
   private static final BiMap<Integer, String> CODE_LABEL = new ImmutableBiMap.Builder<Integer, String>()
       .put(INSERT_CODE, "INSERT")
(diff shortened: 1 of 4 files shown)


commonlib/src/main/java/com/streamsets/pipeline/lib/operation/OperationType.java
```diff
@@ -39,6 +40,7 @@
       .put(UNSUPPORTED_CODE, "UNSUPPORTED")
       .put(UNDELETE_CODE, "UNDELETE")
       .put(REPLACE_CODE, "REPLACE")
+      .put(MERGE_CODE, "MERGE")
       .build();
 
   private static final ImmutableMap<String, Integer> LABEL_CODE = new ImmutableMap.Builder<String, Integer>()
```
ml10: Thoughts on removing `LABEL_CODE` now that `CODE_LABEL` is a `BiMap`?

jarcec: I would be open to it, but let's do that as a separate JIRA to keep the scope of this one limited.

jarcec:
I just merged the change, thank you for your contribution @ml10 !

PR #53, opened 2018-01-30
RabbitMQ Destination - Expiration Property issue - SDC-8371

lazydba247:
change expiration datatype to long,
changed default value to -1 (don't add expiration to message), 
0 Expiration creates auto discarded messages in the queue, 
if no consumers exists listening to the queue


dimaspivak:
I’m not super familiar with this stage but in general, changing default values for a stage could cause compatibility problems for users relying upon them being preset. Our SDK for Python, for example, would probably start misbehaving.

lazydba247:
It is a default for destination, so only effects the rabbitmq? Why should it effect sdk for python?

Not a must to change default, can also add an option to stop/start sending the experation property, the default 0 is causing it to make this stage useless. Without active consumers you lose your data when specifying Persistent as the delivery option.

lazydba247:
updating my pull, removing this, thought of simpler solution

PR #58, opened 2018-03-20
Feature/sdc 4254 excel plugin

EileenSpantree:
### Description

In response to Jira issue, this pull request adds parsing support for Excel, both .xls and .xlsx files.

### Type of Change

New functionality

### Motivation and Context

Several StreamSets uses and commercial customers have expressed an interest in processing XLS and XLSX files in various file origins:

https://issues.streamsets.com/browse/SDC-4254
https://issues.streamsets.com/browse/SDC-5135

### Implementation Details

To provide this feature, we leverage the Apache POI library to parse both XLS and XLSX. Our design patterns drew inspiration from the delimited parser which offers similar functionality for CSVs.

We have added our parser to the following origins:

* Spooldir
* S3
* Google Cloud Storage
* Remote 

### How has this been tested?

For expediency of writing our initial tests, we utilized the Spock framework for terseness. However, we in the process of are translating these tests to JUnit for consistency with the other StreamSets components.

We have also validated against via the SDC UI to confirm functionality (screenshots below)

screenshot  
<br/>
<br/> 
screen shot 2018-03-20 at 10 16 26 am 
<br/>
<br/>
screen shot 2018-03-20 at 10 16 43 am
<br/>

As POI processes both .xls and .xlsx similarly, we don't require the user to specify which version they are using in the UI.  We are currently testing with Spock, and can transition tests to JUnit after you review the current tests.  Here are screenshots of the updated UI.

### Documentation

This feature does not yet have user-facing documentation. Per guidance from StreamSets, Inc. we plan to work with one of StreamSets technical writers to ensure documentation coverage.


commonlib/pom.xml
```diff
@@ -42,6 +42,9 @@
     <joda.version>2.8</joda.version>
     <commons-pool2.version>2.4.2</commons-pool2.version>
     <xerces.version>2.11.0</xerces.version>
+    <poi.version>3.17</poi.version>
+    <groovy.version>2.4.14</groovy.version>
```
jarcec: I wonder why the groovy dependency? Is that just a side effect of depending on Spock?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Cells.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+    switch (cellType) {
+      case STRING:
+        return Field.create(cell.getStringCellValue());
+      case NUMERIC:
+        return Field.create(cell.getNumericCellValue());
+      case BOOLEAN:
+        return Field.create(cell.getBooleanCellValue());
+      default:
+        return null;
```
jarcec: Shouldn't we throw an exception here or something? Otherwise, this could potentially silently loose data, right?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Errors.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License. See accompanying LICENSE file.
+ */
+package com.streamsets.pipeline.lib.parser.excel;
+
+import com.streamsets.pipeline.api.ErrorCode;
+import com.streamsets.pipeline.api.GenerateResourceBundle;
+
+@GenerateResourceBundle
+public enum Errors implements ErrorCode {
+  POI_PARSER_00("Workbooks are only supported as binary data"),
```
jarcec: Really a "nit": Would it make sense to make the error codes "EXCEL_XX" so that it's somehow consistent with the parser itself? So that user knows that they selected Excel as a file format, they see Excel as an exception and make the connection? This way, they might not know that "POI" is our implementation and might not. know where is the error coming from.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Offsets.java
```diff
@@ -0,0 +1,56 @@
...
+    String sheetName = row.getSheet().getSheetName();
+    int rowNum = row.getRowNum();
+    return String.format("%s::%d", sheetName, rowNum);
+  }
+
+  public static Optional<Offset> parse(String offset) {
+    String[] parts = offset.split("::");
+    if (parts.length != 2) {
+      return Optional.empty();
+    }
```
jarcec: I wonder, if these returns let say 3 or 1, then the format is really corrupted and not "empty", right? Would it make more sense to throw an exception and force the user to reset origin (with all the "I'm losing offset" that belongs to it).

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,140 @@
...
+import com.streamsets.pipeline.api.ProtoConfigurableEntity.Context;
+import com.streamsets.pipeline.api.Record;
+import com.streamsets.pipeline.lib.parser.AbstractDataParser;
+import com.streamsets.pipeline.lib.parser.DataParserException;
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.Row;
+import org.apache.poi.ss.usermodel.Workbook;
+
+import java.io.IOException;
+import java.util.*;
```
jarcec: Nit: We usually do not prefer to do asterisk imports, do you mind iterate each of the imported class?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,140 @@
...
+  private boolean eof;
+
+  private List<Field> headers;
+
+  public WorkbookParser(WorkbookParserSettings settings,
+                        Context context,
+                        String readerId,
+                        Workbook workbook,
+                        String offsetId) {
+    this.settings = requireNonNull(settings);
```
jarcec: Nit: I believe that our "new" coding style should be more like this:

```
public WorkbookParser(
    WorkbookParserSettings settings,
     Context context,
    String readerId,
    Workbook workbook,
    String offsetId
) {
```

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,140 @@
...
+    offset = Offsets.offsetOf(currentRow);
+    Record record = context.createRecord(readerId + "::" + offset);
+    LinkedHashMap<String, Field> rowMap = readRow(currentRow);
+    record.set(Field.createListMap(rowMap));
+    return record;
+  }
+
+  @Override
+  public String getOffset() throws DataParserException, IOException {
+    return eof ? "-1" : offset;
```
jarcec: You have to fix me here, I thought that we're expecting null when the parser is done rather then -1?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParserSettings.java
```diff
@@ -0,0 +1,169 @@
...
+public class WorkbookParserSettings {
+  private String startingRow;
+  private String workbookConfiguration;
+  private Map<String, String> sheetSpecificConfiguration;
+  private List<String> sheetsToSkip;
+  private boolean allowExtraColumns;
+  private String extraColumnPrefix;
+  private ExcelHeader header;
+  private int skipStartLines;
+  private int maxObjectLen;
```
jarcec: Since we don't use the Overrun parser class, I believe that we can avoid using the maxObjectLen - or are we using it for slightly different purpose here?

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataFormatParser.java
```diff
@@ -280,6 +281,17 @@ public DataFormatParser(
...
         break;
+      case EXCEL:
+        if (dataFormatConfig.excelHeader == null) {
+          issues.add(
+              context.createConfigIssue(
+                  DataFormatGroups.DATA_FORMAT.name(),
+                  configPrefix + "excelHeader",
+                  DataFormatErrors.DATA_FORMAT_200
+              )
+          );
```
jarcec: This check should not be required - if the given config field is configured as required it should always be non-null (validated by the framework).

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
```diff
@@ -17,35 +17,9 @@
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
-import com.streamsets.pipeline.api.ConfigDef;
-import com.streamsets.pipeline.api.Dependency;
-import com.streamsets.pipeline.api.ListBeanModel;
-import com.streamsets.pipeline.api.ProtoConfigurableEntity;
-import com.streamsets.pipeline.api.Stage;
-import com.streamsets.pipeline.api.ValueChooserModel;
+import com.streamsets.pipeline.api.*;
```
jarcec: Let's back up the unrelated changes.

commonlib/pom.xml
jvz: Yes, that's only included for Spock here.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
jvz: Do you perhaps have IntelliJ or Eclipse format settings we can use?

jarcec: Good question, I think that internally we do,  let me see in the team if someone can export it.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
EileenSpantree: Hi, I believe this is consistent with the `getOffset()` method in the parent `DataParser` class, [which is commented to say](https://github.com/Spantree/streamsets-datacollector/blob/07c1dd23369ad55a30cd039d96751155a7dbfe8b/commonlib/src/main/java/com/streamsets/pipeline/lib/parser/DataParser.java#L29) it returns -1 after reaching the end of the file.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Cells.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+    switch (cellType) {
+      case STRING:
+        return Field.create(cell.getStringCellValue());
+      case NUMERIC:
+        return Field.create(cell.getNumericCellValue());
+      case BOOLEAN:
+        return Field.create(cell.getBooleanCellValue());
+      default:
+        try {
```
jvz: I think you can add this exception to the method signature and let it propagate upwards. The caller of the parser class is responsible for handling the exception.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Offsets.java
```diff
@@ -0,0 +1,64 @@
...
+public class Offsets {
+  public static String offsetOf(Row row) {
+    String sheetName = row.getSheet().getSheetName();
+    int rowNum = row.getRowNum();
+    return String.format("%s::%d", sheetName, rowNum);
+  }
+
+  public static Optional<Offset> parse(String offset) {
+    String[] parts = offset.split("::");
+    if (offset.equals("0")) {
```
jvz: You can do this check before the split above. I'd also suggest adding a line at the top: `if (offset == null) return Optional.empty();`

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Offsets.java
```diff
@@ -0,0 +1,64 @@
...
+      return Optional.empty();
+    }
+    if (parts.length != 2) {
+      try {
+        throw new DataParserException(Errors.EXCEL_PARSER_01);
+      } catch (DataParserException e) {
```
jvz: Like in Cells, this should be added to the method signature to let the exception propagate.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,143 @@
...
+                        String offsetId) {
+    this.settings = requireNonNull(settings);
+    this.context = requireNonNull(context);
+    this.readerId = requireNonNull(readerId);
+    this.workbook = requireNonNull(workbook);
+    this.rowIterator = iterate(this.workbook);
+    this.offset = requireNonNull(offsetId);
+
+    if (!rowIterator.hasNext()) {
+      throw new UnsupportedOperationException("empty workbook");
```
jvz: I'm wondering if we should put this in a `DataParserException` with a localized `Errors` message.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParserConstants.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License. See accompanying LICENSE file.
+ */
+package com.streamsets.pipeline.lib.parser.excel;
+
+public final class WorkbookParserConstants {
+  public static final String STARTING_ROW = "startingRow";
```
jvz: Should we make the unused constants private or remove them?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParserSettings.java
```diff
@@ -0,0 +1,160 @@
...
+ * limitations under the License. See accompanying LICENSE file.
+ */
+package com.streamsets.pipeline.lib.parser.excel;
+
+import com.streamsets.pipeline.config.ExcelHeader;
+
+import java.util.List;
+import java.util.Map;
+
+public class WorkbookParserSettings {
```
jvz: Similarly in this class, should we remove or make private the unsupported options?

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
```diff
@@ -1031,6 +1029,20 @@
   )
   public boolean verifyChecksum = false;
 
+  @ConfigDef(
+          required = true,
+          type = ConfigDef.Type.MODEL,
+          defaultValue = "NO_HEADER",
```
jvz: It might make more sense to go with `WITH_HEADERS` as the default. What do you think?

commonlib/src/test/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParserSpec.groovy
```diff
@@ -0,0 +1,176 @@
...
+
+import com.streamsets.pipeline.api.Field
+import com.streamsets.pipeline.api.OnRecordError
+import com.streamsets.pipeline.api.Stage
+import com.streamsets.pipeline.config.ExcelHeader
+import com.streamsets.pipeline.sdk.ContextInfoCreator
+import org.apache.poi.ss.usermodel.WorkbookFactory
+import spock.lang.Specification
+
+class WorkbookParserSpec extends Specification {
```
jvz: Note to self: if this class is removed, then so can most of the changes in `commonlib/pom.xml` minus the poi dependencies.

commonlib/pom.xml
```diff
@@ -42,6 +42,7 @@
     <joda.version>2.8</joda.version>
     <commons-pool2.version>2.4.2</commons-pool2.version>
     <xerces.version>2.11.0</xerces.version>
+    <poi.version>3.17</poi.version>
```
jarcec: I wonder, have you see any errors reported by our classpath validation code (usually printed in logs during startup) with this new dependency in any stage lib?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/DataParserFormat.java
```diff
@@ -52,7 +53,7 @@
   WHOLE_FILE(WholeFileDataParserFactory.class, WholeFileDataParserFactory.MODES, WholeFileDataParserFactory.CONFIGS),
   SYSLOG(SyslogDataParserFactory.class, SyslogDataParserFactory.MODES, SyslogDataParserFactory.CONFIGS),
   NETFLOW(NetflowDataParserFactory.class, NetflowDataParserFactory.MODES, NetflowDataParserFactory.CONFIGS),
-  ;
+  EXCEL(WorkbookParserFactory.class, WorkbookParserFactory.MODES, WorkbookParserFactory.CONFIGS);
```
jarcec: Let's stick to the original formatting here - let's end this line with a comma and lets the ";" be on its own line. This way when someone will add a future file format they will need to insert only one line and not change existing line. Which is much cleaner for backports and git blame.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Cells.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+    switch (cellType) {
+      case STRING:
+        return Field.create(cell.getStringCellValue());
+      case NUMERIC:
+        return Field.create(cell.getNumericCellValue());
+      case BOOLEAN:
+        return Field.create(cell.getBooleanCellValue());
+      default:
+        throw new DataParserException(Errors.EXCEL_PARSER_01);
```
jarcec: For better debuggability what about improving the error here and including the "cellType" as part of the message? Something like:

throw new DataParserException(~"Unsupported Cell type {}", cellType); (where the string is part of some error defined in Errors of course)

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParserFactory.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+import java.util.Set;
+
+public class WorkbookParserFactory extends DataParserFactory {
+  public static final Set<Class<? extends Enum>> MODES;
+  public static final Map<String, Object> CONFIGS;
+
+  static {
+    CONFIGS = Collections.emptyMap();
+    MODES = ImmutableSet.of(ExcelHeader.class);
```
jarcec: This dependency on guava seems unnecessary - can't we just do Collections.singleton()?

(guava is a 2.2MB large dependency with very bad backward compatibility guarantees, this maven module "commonlib" is inserted into every single stage library so we should be careful about introducing unnecessary dependencies)

commonlib/src/test/java/com/streamsets/pipeline/lib/parser/excel/TestWorkbookParser.java
```diff
@@ -0,0 +1,204 @@
...
+      contentMap.put("column" + i, Field.create((double) i));
+    }
+    Field expected = Field.createListMap(contentMap);
+
+    assertEquals(expected, firstContentRow.get());
+  }
+
+  @Test
+  public void testParseCorrectlyHandlesFileThatIgnoresHeaders() throws IOException, DataParserException, InvalidFormatException {
+    InputStream file = getFile("/TestFile.xlsx");
```
jarcec: I wonder if rat will accept the new binary files - don't we need explicit exclude in the main pom file?

commonlib/pom.xml
EileenSpantree: Just now I started up the data collector locally and the terminal output didn't show any errors related to this dependency.  Is there anywhere else I should check for logs?

jeff303:
It seems that one of the test files has not been added to the repo - `TestFileBlankCells.xlsx`.  I get an NPE when running `com.streamsets.pipeline.lib.parser.excel.TestWorkbookParser#testParseHandlesBlanksCells`.

Also, while we're at it, can we please move all the test resources (Excel files) into a specific package/folder underneath `commonlib/test/resources`, instead of leaving them at the root level?  It will help keep things more organized.  I would recommend simply keeping the same package as each test (ex: `com.streamsets.pipeline.lib.parser.excel` most likely).

Will leave other comments in specific files; thanks!

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Offsets.java
```diff
@@ -0,0 +1,64 @@
...
+  public static Optional<Offset> parse(String offset) {
+    if (offset.equals("0")) {
+      return Optional.empty();
+    }
+    String[] parts = offset.split("::");
+    if (parts.length != 2) {
+      try {
+        throw new DataParserException(Errors.EXCEL_PARSER_01);
+      } catch (DataParserException e) {
+        e.printStackTrace();
```
jeff303: What's going on here?  Is this supposed to catch its own exception?  In any event, if there is an error that needs to be propagated out, this won't do that.  If we simply want to log an error message but still return `Optional.empty()`, then we need to use an SLF4J logger to log that message.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,160 @@
...
+
+  private LinkedHashMap<String, Field> readRow(Row row) throws DataParserException {
+    LinkedHashMap<String, Field> output = new LinkedHashMap<>();
+    String sheetName = row.getSheet().getSheetName();
+    String columnHeader;
+    for (int columnNum = 0; columnNum < row.getLastCellNum(); columnNum++) {
+      if (headers.isEmpty()) {
+        columnHeader = String.valueOf(columnNum);
+      }
+      else {
```
jeff303: Please wrap with previous line

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/Cells.java
```diff
@@ -0,0 +1,43 @@
...
+  static Field parseCell(Cell cell) throws DataParserException {
+    CellType cellType = cell.getCellTypeEnum();
+    if (cell.getCellTypeEnum().equals(CellType.FORMULA)) {
+      cellType = cell.getCachedFormulaResultTypeEnum();
+    }
+    switch (cellType) {
+      case STRING:
+        return Field.create(cell.getStringCellValue());
+      case NUMERIC:
+        return Field.create(cell.getNumericCellValue());
```
jeff303: Without digging too much into the POI code to understand how things are represented... my impression was Excel supported the concept of arbitrary precision decimal values (ex: precision and scale).  How are those represented by POI?  If they are considered NUMERIC, then it seems that there would be slight numeric errors due to double representation here.  Just wanted to confirm this is expected.

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,160 @@
...
+    this.rowIterator = iterate(this.workbook);
+    this.offset = requireNonNull(offsetId);
+
+    if (!rowIterator.hasNext()) {
+      throw new DataParserException(Errors.EXCEL_PARSER_04);
+    }
+
+    switch (settings.getHeader()) {
+      case WITH_HEADER:
+        rowIterator.next();
```
jeff303: Why not do `headerRow = rowIterator.next()` here, instead of advancing the iterator here (and ignoring the result), then fetching via the `sheet.getRow(0)` call below?

commonlib/src/main/java/com/streamsets/pipeline/lib/parser/excel/WorkbookParser.java
```diff
@@ -0,0 +1,160 @@
...
+          headers.put(sheet.getSheetName(), sheetHeaders);
+        }
+
+        break;
+      case IGNORE_HEADER:
+        Row ignored = rowIterator.next();
+        headers = new HashMap<>();
+        break;
+      case NO_HEADER:
+        headers = new HashMap<>();
```
jeff303: Since all clauses do `headers = new HashMap<>()`, why not simply make it final and initialize in the declaration at the top?

pom.xml
```diff
@@ -201,6 +201,7 @@
     <module>spark-executor-protolib</module>
     <module>spark-processor-protolib</module>
     <module>scripting-protolib</module>
+    <module>poi-protolib</module>
```
jeff303: I'm not getting Data Collector to compile because of this (module doesn't exist).  It seems that the new code you wrote is all in `commonlib`, so perhaps this is a spurious module name?

pom.xml
```diff
@@ -514,6 +519,7 @@
         <module>apache-kudu_1_5-lib</module>
         <module>apache-kudu_1_6-lib</module>
         <module>apache-kudu_1_7-lib</module>
+        <module>apache-poi_3_17-lib</module>
```
jeff303: This is another non-existing module that fails in compilation.

commonlib/src/test/java/com/streamsets/pipeline/lib/parser/excel/TestWorkbookParser.java
```diff
@@ -0,0 +1,247 @@
...
+    for (int i = 1; i <= 5; i++) {
+      thirdContentMap.put("header" + i, Field.create((double) i*10));
+    }
+    Field expectedThirdRow = Field.createListMap(thirdContentMap);
+
+    LinkedHashMap<String, Field> fourthContentMap = new LinkedHashMap<>();
+    for (int i = 1; i <= 5; i++) {
+      fourthContentMap.put("header" + i, Field.create((double) i*10));
+    }
+    Field expectedFourthRow = Field.createListMap(fourthContentMap);
```
jeff303: These lines are repeated 4x in this test case.  Would it be possible to create a helper method to construct the `Field` and call that instead?  Looks like it would have a single different parameter to select between the first two cases and the second two.

jeff303:
This has been merged via the internal code review mechanism, under SDC-4254, and can be closed here.

PR #60, opened 2018-05-04
SDC-8903 - Add ability to get confluent subject name from record header

vkrot:
Added dynamic resolution of confluent subject based on record header.


kunickiaj:
Thanks for the contribution! We'll review this shortly.

commonlib/src/main/java/com/streamsets/pipeline/lib/generator/avro/AvroSchemaMetadataProvider.java
```diff
@@ -0,0 +1,97 @@
+package com.streamsets.pipeline.lib.generator.avro;
```
jarcec: Missing license header.

commonlib/src/main/java/com/streamsets/pipeline/stage/destination/lib/DataGeneratorFormatConfig.java
```diff
@@ -375,6 +376,20 @@
...
+      label = "Subject Header",
+      dependencies = {
+          @Dependency(configName = "dataFormat^", triggeredByValues = "AVRO"),
+          @Dependency(configName = "avroSchemaSource", triggeredByValues = "REGISTRY"),
+          @Dependency(configName = "schemaLookupMode", triggeredByValues = "SUBJECT_HEADER")
+      },
+      displayPosition = 452,
+      group = "DATA_FORMAT"
+  )
+  public String subjectHeader;
```
jarcec: I wonder, why not directly adding an ability to put arbitrary EL expression here to get the schema from arbitrary header expression?

vkrot: Good point, I just missed this option. Already applied

vkrot:
Guys, I'm sorry for such a long delay, completely forgot about this PR. I've applied your notes, could you please take a look at it once again?

WgStreamsets:
We are interested in such PR to simplify some pipelines

commonlib/src/main/java/com/streamsets/pipeline/stage/destination/lib/DataGeneratorFormatConfig.java
```diff
@@ -364,14 +364,18 @@
   @ConfigDef(
       required = true,
       type = ConfigDef.Type.STRING,
+      defaultValue = "${record:attribute('subject')}",
```
jarcec: A brainstorming - in this case, there is a performance implication on knowing that the value will never change versus when it could potentially be different for each record. What about rather exposing an explicit checkbox "Retrieve Subject from Record" and provide a different configuration option that will actually explicitly support ELs while keeping this one unchanged. What do you think?

jarcec:
Have you seen my last comment @vkrot ?

commonlib/src/main/java/com/streamsets/pipeline/stage/destination/lib/DataGeneratorFormatConfig.java
vkrot: Sorry, I missed you comment. 
If you have a static subject - just put some static value in EL expression. In this case it will not evaluate it on each record, instead static value will be used. See code branch at AvroDataGeneratorFactory#134.

PR #61, opened 2018-05-14
SDC-8038 Change Oracle CDC Source Config to allow to add more then one pattern for Schemas

JSlowik:
Updated the OracleCDCSource and TableJDBCSource classes to support a comma-separated list of schemas.

### DIFF
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
@@ -1345,7 +1345,7 @@ public class OracleCDCSource extends BaseSource {
   {
     return schema -> {
       SchemaTableConfigBean newBean = new SchemaTableConfigBean();
-      newBean.schema = schema;
+      newBean.schema = schema.trim();
       newBean.excludePattern = baseBean.excludePattern;
       newBean.table = baseBean.table;
       return newBean;
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
@@ -18,6 +18,7 @@ package com.streamsets.pipeline.stage.origin.jdbc.cdc.oracle;
 import com.codahale.metrics.Gauge;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
 import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.streamsets.pipeline.api.BatchMaker;
@@ -186,6 +187,7 @@ public class OracleCDCSource extends BaseSource {
   public static final String STARTED_LOG_MINER_WITH_START_SCN_AND_END_SCN = "Started LogMiner with start SCN: {} and end SCN: {}";
   public static final String REDO_SELECT_QUERY = "Redo select query for selectFromLogMnrContents = {}";
   public static final String CURRENT_LATEST_SCN_IS = "Current latest SCN is: {}";
+  private static final String REGEX_COMMAS_OUTSIDE_QUOTES = ",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)";
 
   public static final int LOGMINER_START_MUST_BE_CALLED = 1306;
   private DateTimeColumnHandler dateTimeColumnHandler;
@@ -213,7 +215,6 @@ public class OracleCDCSource extends BaseSource {
   private ZoneId zoneId;
   private Record dummyRecord;
   private boolean useLocalBuffering;
-
   private Gauge<Map<String, Object>> delay;
   private CallableStatement startLogMnrSCNToDate;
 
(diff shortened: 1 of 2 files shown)


jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -1288,6 +1291,29 @@ private void startLogMinerUsingGivenDates(String startDate, String endDate) thro
     return issues;
   }
 
+  private List<SchemaTableConfigBean> expandListOfSchemas()
```
harishreedharan: This method is actually difficult to understand. I think it might be cleaner if you simply use old-school loops, rather than lambdas (the whole stream / collect pattern makes it pretty verbose anyway).

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -1288,6 +1291,29 @@ private void startLogMinerUsingGivenDates(String startDate, String endDate) thro
...
+  {
+    List<SchemaTableConfigBean> expandedSchemas = new ArrayList<>();
+    configBean.baseConfigBean.schemaTableConfigs
+            .stream()
+            .map(baseConfig -> Arrays.stream(baseConfig.schema.split(","))
```
harishreedharan: if schemas are quoted (the case-sensitive option is selected), then "," is a valid character in a schema name, which could potentially break this logic.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: fair enough, i'll update!

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: hmm... good point, do you have a preferred delimiter?  maybe a pipe?

harishreedharan: I am not sure. Check the oracle docs and choose a character that is not allowed even inside quotes.

harishreedharan: Based on oracle's docs [here](https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm):

`Quoted identifiers can contain any characters and punctuations marks as well as spaces. However, neither quoted nor nonquoted identifiers can contain double quotation marks or the null character (\0).`

That makes it hard 😞

JSlowik: Thanks for the feedback, I'll talk to our lead DBA and see if he has any thoughts.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: @harishreedharan I replaced the lambda with standard loops, and also replaced the basic split on comma with regex that would ignore commas that were within double-quotes.  Thoughts?

harishreedharan: So this would require that any schemas which have `,` character must be specified inside double-quotes, right? That is fair, we can document it. That means we need to strip the double-quote characters though right?

harishreedharan: Do you have any thoughts on validating a bad input like: `"stream, sets` <-- note the unclosed double quotes?

JSlowik: good point on stripping quotes.  I'll run some tests w/ that and push an update in the next day or so. 

Also I'll ponder the bad input question and get back to ya 👍

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -1325,6 +1329,28 @@ private void startLogMinerUsingGivenDates(String startDate, String endDate) thro
     return issues;
   }
 
+  private List<SchemaTableConfigBean> expandListOfSchemas()
+  {
+    List<SchemaTableConfigBean> expandedSchemas = new ArrayList<>();
+    for (SchemaTableConfigBean baseConfig : configBean.baseConfigBean.schemaTableConfigs) {
+      for(String s: Splitter.on(Pattern.compile(REGEX_COMMAS_OUTSIDE_QUOTES)).split(baseConfig.schema)){
```
harishreedharan: Pattern.compile can just be done once - it is an expensive call.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/table/TableJdbcSource.java
```diff
@@ -177,4 +182,35 @@ private void upgradeFromV1() throws StageException {
       offsetKeysToRemove.forEach(tableName -> getContext().commitOffset(tableName, null));
     }
   }
+
+  private List<TableConfigBean> expandAllSchemas()
```
harishreedharan: I like the idea of supporting the same thing here - but don't you need the same regex stuff here too?

JSlowik: Honestly I forgot I had added this section :).

I'll refactor it to match.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: @harishreedharan I've thought about the idea of bad input, and I feel like trying to account for such a thing can lead down a pretty dangerous rabbit hole.  I guess one thing we could look to do is make sure that there are an even number of double quotes in the string otherwise throw an error?

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/table/TableJdbcSource.java
```diff
@@ -183,26 +189,24 @@ private void upgradeFromV1() throws StageException {
...
+    List<TableConfigBean> expandedSchemas = new ArrayList<>();
+    for (TableConfigBean tableConfigBean : tableJdbcConfigBean.tableConfigs) {
+      for(String s: Splitter.on(COMMA_SPLIT_PATTERN).split(tableConfigBean.schema)){
+        TableConfigBean configBean = copyBean(tableConfigBean,s);
+        expandedSchemas.add(configBean);
+      }
+    }
+    return expandedSchemas;
 
   }
```
JSlowik: @harishreedharan I updated TableJdbcSource to match the logic I implemented in SchemaJdbcSource, but have been debating the idea of making the methods a bit more modular and having them both call a utility class.  

Do you have a common class a "splitter" method like this could go under / should I make one ?  or do you want to keep them separate??

harishreedharan:
Hey @JSlowik, sorry I was out on leave since July. Is this ready for review?

jarcec:
It seems that we haven't heard back for several months, so I will go ahead and resolve the pull request. Feel free to submit a new one @JSlowik  when you get to this again!

PR #62, opened 2018-05-29
SDC-9109 Add RUNNING_ERROR state for Notification Triggers.

JSlowik:
Allow triggers to be fired on RUNNING_ERROR state.


JSlowik:
Closed as i screwed up my commit history

PR #64, opened 2018-06-01
Improove mac compatibility

abruneau:
Check if the platform is Darwin and use native Mac SHA1 function if so

### DIFF
--- a/dist/src/main/libexec/_stagelibs
+++ b/dist/src/main/libexec/_stagelibs
@@ -85,11 +85,17 @@ run() {
 # output: NONE, fails if not OK
 #
 verifySHA1() {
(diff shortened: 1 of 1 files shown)


jarcec:
I wonder if you have seen our [BUILD.md](https://github.com/streamsets/datacollector/blob/master/BUILD.md) file @abruneau ? We recommend to `brew install md5sha1sum`. Majority of developers are running on mac so those instructions are actually up to date.

abruneau:
I've seen it. I just wanted to ease the process. It is not a big change but it is one less thing to take care of.

jarcec:
Let's keep this one as is - I'm not too worried about external dependency as this is one of many.

PR #66, opened 2018-06-11
SDC-9202 Allow Oracle CDC recovery option when missing LogFile

JSlowik:
The purpose of this Pull Request is to allow the Oracle CDC Client to "recover" in the event a log file goes missing.  

Specifically, in my use case, our RDS rolls its log file over every 5 minutes.  Therefore, if I stop the pipeline to do some maintenance, and restart after the rollover window has passed, CDC generation fails. (Note, this fix also solves SDC-9006 where no StageException was being logged if LogFile was missing on startup).

This PR adds functionality to allow the plugin to find the first available current Redo Log entry and resume from that position in the event the log file for the current offset is unavailable.

I am open to other approaches to fixing this issue, but hope this can at the minimum spark discussion.

### DIFF
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
@@ -55,9 +55,10 @@ public class OracleCDCConfigBean {
   @ConfigDef(
       required = true,
       type = ConfigDef.Type.BOOLEAN,
-      label = "Missing Log File Recovery",
-      description = "If the log file is removed for the current window, reset to the first available, rather than throwing error",
-      displayPosition = 180,
+      label = "Attempt missing log file recovery",
+      description = "If the log file is removed for the current window, attempt recovery by resetting the offset to " +
+          "the first available time/SCN",
+      displayPosition = 45,
       group = "CDC",
       defaultValue = "false"
   )
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
@@ -57,7 +57,7 @@ public class OracleCDCConfigBean {
       type = ConfigDef.Type.BOOLEAN,
       label = "Attempt missing log file recovery",
       description = "If the log file is removed for the current window, attempt recovery by resetting the offset to " +
-          "the first available time/SCN",
+          "the first available time/SCN of the CURRENT redo log.",
       displayPosition = 45,
       group = "CDC",
       defaultValue = "false"
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
@@ -54,6 +54,17 @@ public class OracleCDCConfigBean {
 
   @ConfigDef(
       required = true,
+      type = ConfigDef.Type.BOOLEAN,
+      label = "Missing Log File Recovery",
+      description = "If the log file is removed for the current window, reset to the first available, rather than throwing error",
+      displayPosition = 180,
+      group = "CDC",
+      defaultValue = "false"
+  )
+  public boolean logFileRecovery;
+
+  @ConfigDef(
+      required = true,
       type = ConfigDef.Type.STRING,
       label = "Start Date",
       description = "Datetime to use for the initial change. Use the following format: DD-MM-YYYY HH24:MM:SS.",
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
@@ -326,6 +326,7 @@ public class OracleCDCSource extends BaseSource {
             throw new StageException(JDBC_86,ex);
           } else {
             offset = firstAvailable;
+            continue;
           }
         }
         LOG.error("Error while attempting to produce records", ex);
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
@@ -317,11 +317,10 @@ public class OracleCDCSource extends BaseSource {
         if (getContext().isPreview() && ex instanceof SQLException) {
           LOG.warn("Exception while previewing", ex);
           return NULL;
-        }
-        else if(ex instanceof SQLException && configBean.logFileRecovery) {
+        } else if (ex instanceof SQLException && configBean.logFileRecovery) {
           LOG.info("Attempting to reset to first valid offset");
           String firstAvailable = resetCDCToFirstAvailable();
-          if(offset.equalsIgnoreCase(firstAvailable)) {
+          if (offset.equalsIgnoreCase(firstAvailable)) {
             LOG.error("No valid redo log available");
             throw new StageException(JDBC_86,ex);
           } else {
@@ -371,7 +370,7 @@ public class OracleCDCSource extends BaseSource {
   private void startGeneratorThread(String lastSourceOffset) throws StageException, SQLException {
     Offset offset = null;
     LocalDateTime startTimestamp;
-    try{
+    try {
       startLogMnrForRedoDict();
       if (!StringUtils.isEmpty(lastSourceOffset)) {
         offset = new Offset(lastSourceOffset);
@@ -411,9 +410,8 @@ public class OracleCDCSource extends BaseSource {
     } catch (SQLException ex) {
       //If the exception is due to a missing log file, bubble it up so possible
       //recovery could be attempted.
-      if(ex.getErrorCode() == MISSING_LOG_FILE || ex.getErrorCode() == NO_LOG_FOUND)
-      {
-        throw new SQLException(ex);
+      if(ex.getErrorCode() == MISSING_LOG_FILE || ex.getErrorCode() == NO_LOG_FOUND) {
+        throw ex;
       }
       LOG.error("SQLException while trying to setup record generator thread", ex);
       generationStarted = false;
@@ -431,9 +429,7 @@ public class OracleCDCSource extends BaseSource {
     });
   }
 
-  private String resetCDCToFirstAvailable()
-  {
-
+  private String resetCDCToFirstAvailable() {
     Offset offset = null;
     try(PreparedStatement ps = connection.prepareStatement(GET_EARLIEST_AVAILABLE_OFFSET);
         ResultSet rs = ps.executeQuery()) {
@@ -449,7 +445,6 @@ public class OracleCDCSource extends BaseSource {
     }
     return offset == null ? "" : offset.toString();
   }
-
 
   @NotNull
   private LocalDateTime adjustStartTimeAndStartLogMnr(LocalDateTime startDate) throws SQLException, StageException {
--- a/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
+++ b/jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
@@ -424,10 +424,8 @@ public class OracleCDCSource extends BaseSource {
 
     Offset offset = null;
     try(PreparedStatement ps = connection.prepareStatement(GET_EARLIEST_AVAILABLE_OFFSET);
-        ResultSet rs = ps.executeQuery())
-    {
-      if(rs.next())
-      {
+        ResultSet rs = ps.executeQuery()) {
+      if(rs.next()) {
         Timestamp t = rs.getTimestamp("FIRST_TIME");
         BigDecimal firstSCN = rs.getBigDecimal("FIRST_CHANGE");
         offset = new Offset(version, t.toLocalDateTime(), firstSCN.toPlainString(), 0);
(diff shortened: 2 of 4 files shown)


jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +316,18 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
           LOG.warn("Exception while previewing", ex);
           return NULL;
         }
+        else if(ex instanceof SQLException && ((SQLException)ex).getErrorCode() == MISSING_LOG_FILE && configBean.logFileRecovery)
```
harishreedharan: I think MISSING_LOG_FILE is 1291. I have seen 1292 also pop up (No Log file found for this session)

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +316,18 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
...
+        {
```
harishreedharan: Should be in same line as above.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +316,18 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
...
+          LOG.info("Attempting to reset to first valid offset");
+          String firstAvailable = resetCDCToFirstAvailable();
+          if(offset.equalsIgnoreCase(firstAvailable)) {
+            LOG.error("No valid redo log available");
+            throw new StageException(JDBC_86,ex);
+          }
+          else {
```
harishreedharan: Move this to line above.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +316,18 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
...
+            offset = firstAvailable;
```
harishreedharan: This does not help no? `offset` is a method-level variable, so this assignment is not useful since it will get overwritten by the lastSourceOffset variable when the method is called again. I think you'd need to reattempt staring the generator thread.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -355,47 +370,41 @@ private void pollForStageExceptions() throws StageException {
...
-            throw new StageException(JDBC_82);
-          }
-          startTimestamp = offset.timestamp.minusSeconds(configBean.txnWindow);
-        } else {
-          if (useLocalBuffering) {
-            throw new StageException(JDBC_83);
-          }
-          startTimestamp = getDateForSCN(new BigDecimal(offset.scn));
-          offset.timestamp = startTimestamp;
+    startLogMnrForRedoDict();
```
harishreedharan: Let;s bubble the exception out only for the specific condition where logfiles are missing. I'd like anything else to reattempt a startup of logminer.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: @harishreedharan 

Unless I'm misunderstanding the method, it looks like the "produce" call will loop for "MAX_GENERATION_ATTEMPTS" unless a record gets produced or the generator is stopped (line 298).  Therefore if the CDC generation fails due to a missing logfile, the offset will be adjusted, and on the next loop iteration, this "valid" offset will be utilized.  As long as it receives records then (which it should since that is what LogMiner is reporting as valid), the record will properly set "nextOffset" and return.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: Done, added in both locations

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -223,6 +225,7 @@
     UNKNOWN
   }
 
+  private static final String GET_EARLIEST_AVAILABLE_OFFSET = "SELECT min(FIRST_TIME) as FIRST_TIME, min(FIRST_CHANGE#) as FIRST_CHANGE FROM V$LOG where STATUS = 'CURRENT'";
```
harishreedharan: use STATUS = 'ACTIVE', that way we would be able to get data from files that are not necessarily the latest but are still around

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -393,6 +407,12 @@ private void startGeneratorThread(String lastSourceOffset) throws StageException
         }
       }
     } catch (SQLException ex) {
+      //If the exception is due to a missing log file, bubble it up so possible
+      //recovery could be attempted.
+      if(ex.getErrorCode() == MISSING_LOG_FILE || ex.getErrorCode() == NO_LOG_FOUND)
+      {
+        throw new SQLException(ex);
```
harishreedharan: throw ex is fine, otherwise I am not sure the error code and what not gets set correctly

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +317,16 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
...
           return NULL;
         }
+        else if(ex instanceof SQLException && configBean.logFileRecovery) {
+          LOG.info("Attempting to reset to first valid offset");
+          String firstAvailable = resetCDCToFirstAvailable();
+          if(offset.equalsIgnoreCase(firstAvailable)) {
+            LOG.error("No valid redo log available");
+            throw new StageException(JDBC_86,ex);
+          } else {
+            offset = firstAvailable;
```
harishreedharan: We need a `continue;` here to short-circuit the rest of the loop, so the thread start is attempted again, and we don't directly go to the error handler.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: I initially attempted to use ACTIVE, but that still throws the error.  Granted this is firs time diving into LogMiner, so I could be missing something.  This blog post is where I first saw the suggestion to use CURRENT:  https://oraclespin.com/2010/04/19/example-of-using-logminer/

JSlowik: Digging into it more, it looks like the "ACTIVE" logs are already archived, and they'd have to be manually added to the logminer.  I'm investigating this on my AWS RDS now.

harishreedharan: `CURRENT` works too. Let's make sure we clearly document it in the config description.

JSlowik: Okay sounds good.  I'll add it to the description on the Config Bean.  

I do think there's ways to manually add archived logs to the miner, however AWS does not provide access to them easily.  So unfortunately I cannot really delve into it.  May be worth investigating as a future enhancement though.  

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.Oracle.CommonDBATasks.Log.html

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: good point.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
```diff
@@ -52,6 +52,18 @@
   @ValueChooserModel(StartChooserValues.class)
   public StartValues startValue;
 
+  @ConfigDef(
```
harishreedharan: This requires bumping the schema of the origin up and adding upgrader support to automatically insert this value on upgrades.

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +317,17 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
           LOG.warn("Exception while previewing", ex);
           return NULL;
         }
+        else if(ex instanceof SQLException && configBean.logFileRecovery) {
```
harishreedharan: Style: Should be on previous line. Also space after `if`

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
harishreedharan: `throw ex`

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -409,6 +430,26 @@ private void startGeneratorThread(String lastSourceOffset) throws StageException
     });
   }
 
+  private String resetCDCToFirstAvailable()
+  {
```
harishreedharan: should be on previous line

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -393,6 +408,12 @@ private void startGeneratorThread(String lastSourceOffset) throws StageException
         }
       }
     } catch (SQLException ex) {
+      //If the exception is due to a missing log file, bubble it up so possible
+      //recovery could be attempted.
+      if(ex.getErrorCode() == MISSING_LOG_FILE || ex.getErrorCode() == NO_LOG_FOUND)
+      {
```
harishreedharan: should be on previous line

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -355,7 +370,7 @@ private void pollForStageExceptions() throws StageException {
   private void startGeneratorThread(String lastSourceOffset) throws StageException, SQLException {
     Offset offset = null;
     LocalDateTime startTimestamp;
-    try {
+    try{
```
harishreedharan: nit: space required after `try`

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -313,6 +317,17 @@ public String produce(String lastSourceOffset, int maxBatchSize, final BatchMake
           LOG.warn("Exception while previewing", ex);
           return NULL;
         }
+        else if(ex instanceof SQLException && configBean.logFileRecovery) {
+          LOG.info("Attempting to reset to first valid offset");
+          String firstAvailable = resetCDCToFirstAvailable();
+          if(offset.equalsIgnoreCase(firstAvailable)) {
```
harishreedharan: nit: space after `if`

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
```diff
@@ -409,6 +430,26 @@ private void startGeneratorThread(String lastSourceOffset) throws StageException
...
+      }
+    }
+    catch(SQLException ex){
+      LOG.warn("Could not determine oldest available redo log time");
+      addToStageExceptionsQueue(new StageException(JDBC_02, ex));
+    }
+    return offset == null ? "" : offset.toString();
+  }
+
+
```
harishreedharan: only 1 empty line required here

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCConfigBean.java
JSlowik: @harishreedharan thanks for the input.  was unaware of the upgrader and its usage.  Will address that asap

jdbc-lib/src/main/java/com/streamsets/pipeline/stage/origin/jdbc/cdc/oracle/OracleCDCSource.java
JSlowik: Sorry :)  my work's style is on new line and my formatter adjusted as such automatically!

harishreedharan:
Hey @JSlowik, same question as the other PR, time to review this as well?

jarcec:
I will resolve the pull request for now since we haven't heard back for several months. Feel free to reopen the pull request when you get chance to get back to it @JSlowik !

PR #68, opened 2018-06-20
[SDC-9317] - added delete bin functionality

zlosim:
Hi
added the possibility to delete bins in aerospike from streamsets
please let me know what you think

### DIFF
--- a/basic-lib/src/main/java/com/streamsets/pipeline/lib/microservice/ResponseConfigBean.java
+++ b/basic-lib/src/main/java/com/streamsets/pipeline/lib/microservice/ResponseConfigBean.java
@@ -36,6 +36,6 @@ public class ResponseConfigBean {
   public DataFormat dataFormat = DataFormat.JSON;
 
   @ConfigDefBean(groups = {"#0"})
-  public DataGeneratorFormatConfig dataGeneratorFormatConfig;
+  public DataGeneratorFormatConfig dataGeneratorFormatConfig = new DataGeneratorFormatConfig();
 
 }
--- a/basic-lib/src/main/java/com/streamsets/pipeline/lib/websocket/WebSocketCommon.java
+++ b/basic-lib/src/main/java/com/streamsets/pipeline/lib/websocket/WebSocketCommon.java
@@ -16,10 +16,27 @@
 
 package com.streamsets.pipeline.lib.websocket;
 
+import com.streamsets.pipeline.api.PushSource;
+import com.streamsets.pipeline.api.Record;
+import com.streamsets.pipeline.lib.generator.DataGenerator;
+import com.streamsets.pipeline.lib.generator.DataGeneratorException;
+import com.streamsets.pipeline.lib.generator.DataGeneratorFactory;
+import com.streamsets.pipeline.lib.parser.DataParserFactory;
 import com.streamsets.pipeline.lib.tls.TlsConfigBean;
+import com.streamsets.pipeline.stage.origin.restservice.RestServiceReceiver;
+import com.streamsets.pipeline.stage.util.http.HttpStageUtil;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.jetty.websocket.api.Session;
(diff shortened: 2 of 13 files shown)


jarcec:
I wonder - rather than having an explicit list of deletes, would it make sense to support the CDC-type like functionality that we have for example on MongoDB Destination or JDBC Destination?  (e.g. have tag in the header describing if given record is insert or delete and perform given operation on the destination)

zlosim:
I was thinking about that a lot, but in aerospike destination the action is not per record, but per bin e.g. one record can create mutiple bins and also delete multiple bins in one call to aerospike. So we would probably need a map of action per bin which I think is not very nice solution.

zlosim:
@jarcec what do you think?

jarcec:
We already accept one BinList right? Which specifics which "bins" should be mutated by all records - please correct me if my understanding is wrong here as I need to depend on your Aerospike expertise here.

If that is the case, why not keep that single list and based on the record operation that is consistent with other CDC origins and either perform an insert or delete to those single list of bins?

zlosim:
Just to clarify, if I have single record R1 and want to insert bin B1 and delete bin B2, in Aerospike I can do this with single call and would like to take advantage of this, thus two bin lists. As far as I know, this is not possible in MySQL, Mongo and others.

I can create a single list, add a flag if given bin should be inserted or deleted, show/hide appropriate fields for that particular bin and operation, but I can`t have simple flag in header saying every bin in a list should be deleted XOR inserted.

Sorry if this was your original suggestion and I misinterpreted it.

zlosim:
Can`t really tell for the others, but I can imagine such a scenario where one would like to make as few requests as possible. 
Mine is keeping a directional graph with weights up-to-date - when node is changed, few edges are added and few are deleted, and I`m able to reflect this with a single call from sdc.

My node is represented as a map, (where keys are bins, and values are values in Aerospike terms). So when node is changed, single record (row in aerospike) is updated - few keys are removed from the map, and few are added.
Keeping every node in a map with a flag for connection is out of the question because of memory consumption. Would like to keep the logic in SDC, would like to do as few calls to Aerospike as possible.

jarcec:
My apologies @zlosim , but I'm still unsure how could a typical StreamSet user take advantage of the functionality if we would implemented it in the proposed way.  As far as I know, no existing origin will create the "half insert / half remove" record. 

Do you perhaps have few pipeline ideas how this could be used?

zlosim:
Pretty much any enable/disable functionality where I don`t want to waste memory - I can explicitly say that Feature A is enabled/disabled and waste some memory or I can ask if key for feature A exists or not . GDPR opt ins/outs for example.

jarcec:
What origin would such pipeline use though @zlosim ? Perhaps it's obvious, but I can only imagine Dev Raw Data Origin with specifically crafted data. It's hard for me to think about a whole use case that will read from somewhere where the records will be half insert and half delete as that is not really how we designed the Data Collector.

zlosim:
For me specifically, these are web generated events streaming to my application. Events are generated by multiple "customers" and I`m keeping the business logic inside sdc - thus crafting the data in event evaluator based on the input record. And yes, I can understand this might be very specific use case, so it might be more useful to keep this processor in line with others. Wanted to take advantage of insert-delete approach but can rewrite it as I still would like to see delete functionality in SDC.

zlosim:
Hi @jarcec 
I just reworked the Aerospike target, got rid of the other list and added flag to indicate if its going to be UPDATE or DELETE operation.
please let me know if this is good or need some more work :)
thanks

zlosim:
Hi @jarcec  I know its been a while, but would love to see this in production :) let me know if it needs some more work

jarcec:
Sorry for the delays @zlosim . I appreciate your openness to update the patch to keep consistency with other components. Most of the CDC-like components are using record headers to indicate an operation that should be performed with the record:

https://github.com/streamsets/datacollector/blob/master/stagesupport/src/main/java/com/streamsets/pipeline/lib/operation/OperationType.java

What about using this operation to indicate DELETE or INSERT with having a configuration to select the default operation? This is how all other components work today.

PR #71, opened 2018-07-10
SDC-9414. Non-Resolvable parent POM for wholefile-converter-protolib

eschrock:
Not sure if I should be using SDC-9414 since it was already closed as resolved, but please educate me on the right commit message. Prior to this change, I could not build the project due to the error referenced in that issue:

`[WARNING] 'parent.relativePath' of POM com.streamsets:streamsets-datacollector-wholefile-converter-protolib:3.4.0-SNAPSHOT (/path/to/datacollector/wholefile-converter-protolib/pom.xml) points at com.streamsets:streamsets-datacollector instead of com.streamsets:streamsets-datacollector-root, please verify your project structure @ line 23, column 11`

After this change, I am able to build it successfully.

### DIFF
--- a/wholefile-converter-protolib/pom.xml
+++ b/wholefile-converter-protolib/pom.xml
@@ -22,8 +22,9 @@
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>com.streamsets</groupId>
-    <artifactId>streamsets-datacollector-root</artifactId>
+    <artifactId>streamsets-datacollector-root-lib</artifactId>
     <version>3.4.0-SNAPSHOT</version>
+    <relativePath>../root-lib</relativePath>
   </parent>
   <artifactId>streamsets-datacollector-wholefile-converter-protolib</artifactId>
   <version>3.4.0-SNAPSHOT</version>


wholefile-converter-protolib/pom.xml
```diff
@@ -24,6 +24,7 @@
     <groupId>com.streamsets</groupId>
     <artifactId>streamsets-datacollector-root</artifactId>
     <version>3.4.0-SNAPSHOT</version>
+    <relativePath>../root</relativePath>
```
jarcec: Do you mind changing the inheritance to follow what we do for other protolibs?

Like here: https://github.com/streamsets/datacollector/blob/master/cassandra-protolib/pom.xml

I'm happy to merge it then on the account of being consistent. I'm still not sure why most environments are able to build it properly, but that shouldn't block the patch itself.

eschrock: Sorry, signed the ICLA but forgot to confirm email, should be done now. Will update to follow the root-lib model of others.

eschrock:
Updated to use root-lib to follow other protolibs. Built successfully:

`[INFO] StreamSets Data Collector Whole File Converter ProtoLib SUCCESS [  0.403 s]`

jarcec:
My apologies for the delay @eschrock , I was sidetracked yesterday. Let me submit the patch to our CI system and merge it if it passes.

jarcec:
Thanks for the contribution @eschrock !

PR #72, opened 2018-07-19
Update CredentialStores.html to fix credential expression

tdonovic:
Copy/pasting the vault credential:getwithOptions from the bottom of the page https://streamsets.com/documentation/datacollector/latest/help/#datacollector/UserGuide/Configuration/CredentialStores.html#concept_sbq_vps_51b had a lower case w, when it should be upper case, causing the function to fail. I have fixed this in the docs


jarcec:
Thank you for the report @tdonovic ! The html files are actually generated, so if I would accept the change as-is it would be overridden on the next release. I've followed up with our documentation team to fix it in the source material.

PR #75, opened 2018-08-31
Trivial: Fixing typos in a number of JDBC Origins

salbat:
The typo can noticed in the origin's tooltip for:

- MS SQL Server CDC Origin
- MS SQL Server CT Origin
- Oracle CDC Origin
- PostgreSQL CDC Origin

Also, added "MS" to the MS SQL Server CT Origin to make it consistent with the MS SQL Server CDC Origin.


jarcec:
Good catch @salbat ! It seems that we have fixed the typos before we got a chance to incorporate the pull request, so I will go ahead and close it.

PR #79, opened 2018-12-05
Extensible offset storage, adds possibility to store pipeline offsets…

vkrot:
… in external storage, not only in file. As for now supports only storing offsets in files and in zookeeper. Offset storage is configured for whole sdc instance, not per-pipeline.

### DIFF
--- a/container/pom.xml
+++ b/container/pom.xml
@@ -439,6 +439,25 @@
       <version>0.7.2</version>
     </dependency>
 
+    <!--TODO reviewer, what versions of zookeeper,curator are compatible with sdc? there may be conflicts with other libs I'm not aware of -->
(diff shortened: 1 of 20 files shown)


vkrot:
Guys, any ideas about this PR?

WgStreamsets:
We are interested in such functionality too.

jarcec:
Thank you for the contribution @vkrot . Sadly, we currently do not have plans to make the offset storage pluggable as that would interfere with our future plans for the data collector.

jarcec:
Making the storage module pluggable doesn't really fit our plans with data collector going forward, so I will sadly have to reject the change. I will go ahead and resolve the pull request for now.

PR #80, opened 2018-12-15
This commit fixes the build errors, if built with HEAD of datacollector-api

coolant9:
The problem was in some test files that did not override all the
abstract methods of interceptor. Introduced in commit e1acbafe4be999c90ef8bf98021174af5b994ee3
in datacollector-api


coolant9:
This was addressed shortly by a revert in datacollector-api

PR #82, opened 2019-01-17
SDC-10772. Support durable subscribers in JMS source

mackenza:
added support for durable topic subscriptions in JMS source. Added necessary config to support the JMS API `createDurableSubscriber` (a flag to indicate it's durable and a subscription name).


mackenza:
my commit was screwed up... will try again

PR #83, opened 2019-01-17
SDC-10772. Support durable subscribers in JMS source

mackenza:
Added support for durable topic subscriptions in JMS source. 
Added necessary config to support the JMS API createDurableSubscriber (a flag to indicate it's durable and a subscription name).


mackenza:
@jarcec I didn't include an integration test for 2 reasons.

1. I have no experience with Python. Not a big deal as I could likely figure it out from the existing tests there.
2. I looked at the tests in https://github.com/streamsets/datacollector-tests/blob/master/stage/test_jms_stages.py and it was only a consumer test (on a queue) and a producer test. Since this PR would not introduce a variation on what was already being tested (for instance a "topic" or "unknown" endpoint isn't being tested) I let number 1 above be my guide ;)

jarcec:
The test framework is actually easy to use and deep knowledge if python is definitely not expected. Our documentation is here: https://streamsets.com/documentation/stf/latest/

I would love to have a test that will ensure that the durable subscriber is indeed in use. E.g. something like start pipeline, consume messages, stop pipeline, generate messages, start pipeline, ensure that those messages were read.

mackenza:
Is this a separate PR to thay test repo?

jarcec:
Yeah that would be a separate pull request - for the time being we are keeping both repositories separate, even though we do have plans to merge them eventually.

mackenza:
> Activation key
> 
> The StreamSets Test Framework depends upon the presence of a StreamSets Toolkit activation key in the user’s home folder, as described in the StreamSets SDK for Python documentation. This key is available to customers as part of the StreamSets Premium and Enterprise subscription tier.
I have no subscription.

jarcec:
Let me figure out what I can do for you @mackenza .

mackenza:
ok, so I now have a license key for STF. I installed STF and successfully ran:

`stf test --sdc-version 3.4.0 -s stage/test_dev_stages.py`

Now I want to run `stf test --sdc-version 3.4.0 -s stage/test_jms_stages.py` but I have no success with that. I have ActiveMQ running locally (though in a docker container) on same host.

```
2019-03-04 08:59:40 PM [INFO] [streamsets.testframework.cli] Pulling Docker image streamsets/testframework:master ...
============================= test session starts ==============================
platform linux -- Python 3.6.8, pytest-3.8.2, py-1.8.0, pluggy-0.8.1
benchmark: 3.1.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /root/tests, inifile:
plugins: benchmark-3.1.1
collected 4 items

stage/test_jms_stages.py ssss

========================== 4 skipped in 0.01 seconds ===========================
```
any idea how I get further so that I can actually run my new tests?

kirtiv1:
Hey @mackenza, 
Thanks for adding more functionality !!

Now regarding tests, first thing is we need to see why the tests are being skipped. And for that here is a way: `stf test --sdc-version 3.4.0 -rs stage/test_jms_stages.py`

When I did run that, here is what it said at end: 
```
stage/test_jms_stages.py ssss                                            [100%]
=========================== short test summary info ============================
SKIP [4] /root/testframework/streamsets/testframework/conftest.py:403: JMS tests only run if --jms-server-url is passed
```

And so you need to pass that argument.

kirtiv1:
Then I ran the tests using following:
`test --sdc-version 3.4.0 -vs --jms-server-url tcp://activemq:61616 --jms-client-port 61613 stage/test_jms_stages.py`

when my docker container (with name=activemq) for JMS is set up like following:
`$ docker ps
CONTAINER ID        IMAGE                                              COMMAND                  CREATED             STATUS              PORTS                                                                                                                                                     NAMES
ad6a7b15b65c        webcenter/activemq:5.14.3                          "/app/run.sh"            2 minutes ago       Up 2 minutes        0.0.0.0:32859->1883/tcp, 0.0.0.0:32858->5672/tcp, 0.0.0.0:32857->8161/tcp, 0.0.0.0:32856->61613/tcp, 0.0.0.0:32855->61614/tcp, 0.0.0.0:32854->61616/tcp   activemq`

mackenza:
ok I can try that. I pasted the complete message from my run of STF so not sure why you got the skip information about missing parameters and I did not.

Can you give me the `docker run` you used to run that ActiveMQ on the `cluster` network?

kirtiv1:
why you got the skip information about missing parameters and I did not.
--> The difference is in the command. I gave `-rs` and you did with just `-s`. That -r show extra test summary info as specified by chars (f)ailed, (E)error, (s)skipped, (x)failed, (X)passed,(p)passed, (P)passed with output, (a)all except pP.

Can you give me the docker run you used to run that ActiveMQ on the cluster network?
--> Here it is: 
`docker run --name='activemq' -it --rm --net=cluster  -e 'ACTIVEMQ_MIN_MEMORY=512' -e 'ACTIVEMQ_MAX_MEMORY=2048' -dP webcenter/activemq:5.14.3`

mackenza:
thanks a lot @kirtiv1 , I am now able to run the JMS tests successfully.

I do get this error, though, despite the tests seeming to run:

```
2019-03-05 12:35:52 AM [DEBUG] [streamsets.testframework.sdc] Image (streamsets/sdc-extra-libs:activemq) not found locally. Attempting to pull it ...
Error response from daemon: pull access denied for streamsets/sdc-extra-libs, repository does not exist or may require 'docker login'
2019-03-05 12:35:52 AM [WARNING] [streamsets.testframework.sdc] Error encountered while pulling image (streamsets/sdc-extra-libs:activemq) as None
```
Not sure if that is something you might want to fix or not.

Now I need to write these tests ;)

kirtiv1:
That is great work @mackenza !!

Yes exactly as you discovered we have these docker images built and available and so the tests pass successfully without those errors.
 Now looking forward to your tests !!

mackenza:
@jarcec it looks like I am going to have to add a `clientID` property to make this work correctly.

When I tested it against the Solace PubSub+ broker, it worked fine because Solace JMS client libraries add a clientID automatically as needed. When I tried it on ActiveMQ just now, it needs a `clientID` added as a custom property to work.

I am a little torn about how to handle this. Consider the following:

1. clientID is a property of the connection, not the subscriber/consumer.
1. a clientID can be set in the ConnectionFactory. On ActiveMQ, the default `ConnectionFactory` does not have it set, but if the user was using their own ConnectionFactory, it could be set there.
1. a clientID isn't *only* for durable topic subscribers, but it i*is* required for one. In general, setting the clientID isn't a best practice for shared subscriptions.

So given all that, my thought was that I would make clientID a generic JMS property and not one that is active based on my toggle of `durable` or not.

My question to you is:

1. should it be a normal, not required field or
2. `dependsOn = "durableSubscription"` true? which means you can *only* set it with durable subs?

mackenza:
I added clientID as a general JMS option that is added to the ConnectionFactory. It is independent of the durable topic boolean but this is consistent with the way other JMS clients are configured. This should probably be noted in the docs.

jms-lib/src/main/java/com/streamsets/pipeline/lib/jms/config/BaseJmsConfig.java
```diff
@@ -64,6 +64,16 @@
...
 
+  @ConfigDef(
+      required = false,
+      type = ConfigDef.Type.STRING,
+      label = "Client ID",
+      description = "If specified, must be a unique string used by a single JMS connection instance. Required for Durable Topic Subscriptions",
+      displayPosition = 75,
+      group = "JMS"
+  )
+  public String clientID = null;
```
jarcec: I'm brainstorming here, so bear with me. What if we would add a checkbox "Use Client ID" that would enable or disable this config? Then for JMS systems that do require it the user can primitively select the checkbox and specify it.

And this can be independent on the durable subscriber since as you're saying, not all systems require the client id to be explicitly specified for durable subscriber either.

mackenza: I think that could work. 

To be clear, though, the client ID is not necessarily required for future connections, but in order for the client to use the durable subscription and get messages that were published while they were offline, the `clientID` *must* be the same as it was when the subscription was created. So leaving it to auto-generated client IDs is kind of risky.

mackenza: I will play with a checkbox for clientID later today

mackenza: https://memegenerator.net/img/instances/65418174/just-when-i-thought-i-was-outthey-pull-me-back-in.jpg

jarcec: :-) I submitted the patch to our internal Gerrit to see if things will go well and will wait on the checkbox to actually merge.

mackenza: @jarcec what is the process for upgrade? For instance, if I say that the Boolean for `use clientID` or `durable subscription` are `required`, when I use the new stage library version with an existing pipeline, I get an error that says the required field has no value. Is this how you have done additive properties in the past? Or should the Boolean values be `required = false`?

jarcec: You need to bump the version of the stage here:

https://github.com/streamsets/datacollector/blob/master/jms-lib/src/main/java/com/streamsets/pipeline/stage/origin/jms/JmsDSource.java#L36

And create proper upgrader here:

https://github.com/streamsets/datacollector/blob/master/jms-lib/src/main/java/com/streamsets/pipeline/stage/origin/jms/JmsSourceUpgrader.java#L29

mackenza: I am going to be an old man before this PR is done...

jarcec: We're almost there :)

mackenza: so for the upgrader, I just pick the next increment? so 7?

jarcec: Correct and then upgrade from 6 to 7 (e.g. adding all the new configuration options).

mackenza:
```suggestion
switch(context.getFromVersion()) {
      case 1:
        upgradeV1ToV2(configs);
        if (context.getToVersion() == 2) {
          break;
        }
        // fall through
      case 2:
        upgradeV2ToV3(configs);
        if (context.getToVersion() == 3) {
          break;
        }
        // fall through
      case 3:
        upgradeV3ToV4(configs);
        if (context.getToVersion() == 4) {
          break;
        }
        // fall through
      case 4:
        if (context.getToVersion() == 5) {
          break;
        }
        upgradeV4ToV5(configs);
      // fall through
      case 5:
        upgradeV5ToV6(configs, context);
        break;
      default:
        throw new IllegalStateException(Utils.format("Unexpected fromVersion {}", context.getFromVersion()));
}
```
why in `case 4:` is the test for the `getToVersion` done first and then the upgrade function vs. all the others where the upgrade function is done and then the `getToVersion` is checked?

jarcec: I believe that that is a bug @mackenza , and a fairly old one. The case 4 is correct, the others not so much.

mackenza: am I missing something here? In `case 4` if I am upgrading from 4 (and thus match the case) to 5, it's going to `break;` and not do the `upgradeV4ToV5()`.  Again, I may be missing something. 

The only reason I ask is I need to add something like that for my `upgradeV6ToV7()` I need to write and I need to alter the `case 5:` to do that fall through thing, don't I?

jarcec: Yeah we should upgrade and then break, sorry.

mackenza: ok I added the upgrade stuff and also the boolean for entering clientID. I think it's all set now.

jarcec: Thanks @mackenza ! Let me upload the latest variant to gerrit for automated testing and will merge if successful!

jarcec:
The pull request has been merged, thank you for your contribution @mackenza !

PR #88, opened 2019-03-03
SDC-10501 : Option to enable/disable Kafka auto commit Offsets

SandishKumarHN:
JIRA: SDC-10501 Kafka Multi-Topic Consumer not checking delivery guarantee

Toggle Option for Kafka enable.auto.commit.

- The consumer can either automatically commit offsets periodically, default enable.auto.commit=true, with auto.commit.interval.ms=1000 interval. 
- Or it can choose to control this committed position manually by calling one of the commit commitSync after finishing processing messages.

### DIFF
--- a/kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
+++ b/kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
@@ -17,6 +17,7 @@ package com.streamsets.pipeline.stage.origin.multikafka;
 
 import com.google.common.base.Throwables;
 import com.streamsets.pipeline.api.BatchContext;
+import com.streamsets.pipeline.api.DeliveryGuarantee;
 import com.streamsets.pipeline.api.Field;
 import com.streamsets.pipeline.api.Record;
 import com.streamsets.pipeline.api.Stage;
@@ -116,11 +117,17 @@ public class MultiKafkaSource extends BasePushSource {
             for (ConsumerRecord<String, byte[]> item : messages) {
               list.add(item);
               if (list.size() == conf.maxBatchSize) {
+                if (getContext().getDeliveryGuarantee() == DeliveryGuarantee.AT_MOST_ONCE) {
+                  consumer.commitSync();
+                }
                 sendBatch(list);
                 list.clear();
               }
             }
             messagesProcessed += messages.count();
+            if (getContext().getDeliveryGuarantee() == DeliveryGuarantee.AT_LEAST_ONCE) {
+              consumer.commitSync();
+            }
             LOG.trace("Kafka thread {} finished processing {} messages", this.threadID, messages.count());
           } else {
             if (!list.isEmpty()) {
@@ -313,11 +320,10 @@ public class MultiKafkaSource extends BasePushSource {
     props.setProperty("bootstrap.servers", conf.brokerURI);
     props.setProperty("group.id", conf.consumerGroup);
     props.setProperty("max.poll.records", String.valueOf(batchSize));
-    props.setProperty("enable.auto.commit", "true");
-    props.setProperty("auto.commit.interval.ms", "1000");
     props.setProperty(KafkaConstants.KEY_DESERIALIZER_CLASS_CONFIG, conf.keyDeserializer.getKeyClass());
     props.setProperty(KafkaConstants.VALUE_DESERIALIZER_CLASS_CONFIG, conf.valueDeserializer.getValueClass());
     props.setProperty(KafkaConstants.CONFLUENT_SCHEMA_REGISTRY_URL_CONFIG, StringUtils.join(conf.dataFormatConfig.schemaRegistryUrls, ","));
+    props.setProperty(KafkaConstants.AUTO_COMMIT_OFFEST, "false");
 
     if(context.isPreview()) {
       props.setProperty(KafkaConstants.AUTO_OFFSET_RESET_CONFIG, KafkaConstants.AUTO_OFFSET_RESET_PREVIEW_VALUE);
--- a/kafka-common/src/main/java/com/streamsets/pipeline/lib/kafka/KafkaConstants.java
+++ b/kafka-common/src/main/java/com/streamsets/pipeline/lib/kafka/KafkaConstants.java
@@ -25,5 +25,6 @@ public class KafkaConstants {
   // Constants specific for new consumer/producer library
   public static final String AUTO_OFFSET_RESET_CONFIG = "auto.offset.reset";
   public static final String AUTO_OFFSET_RESET_PREVIEW_VALUE = "earliest";
+  public static final String AUTO_COMMIT_OFFEST = "enable.auto.commit";
 
 }
--- a/kafka_multisource-0_10-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/v0_10/loader/Kafka0_10ConsumerLoader.java
+++ b/kafka_multisource-0_10-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/v0_10/loader/Kafka0_10ConsumerLoader.java
@@ -173,5 +173,8 @@ public class Kafka0_10ConsumerLoader extends KafkaConsumerLoader {
     public void close() {
       delegate.close();
     }
+
+    @Override
+    public void commitSync() { delegate.commitSync(); }
   }
 }
--- a/kafka_multisource-0_9-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/v0_9/loader/Kafka0_9ConsumerLoader.java
+++ b/kafka_multisource-0_9-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/v0_9/loader/Kafka0_9ConsumerLoader.java
@@ -82,5 +82,8 @@ public class Kafka0_9ConsumerLoader extends KafkaConsumerLoader {
     public void close() {
       delegate.close();
     }
+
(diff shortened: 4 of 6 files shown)


brockn:
This looks good. I wonder if an upgrader is needed?

brockn:
I also wonder if instead of making this an option, we just remove this entirely. At the pipeline level you can pick at least once or at most once.

SandishKumarHN:
> This looks good. I wonder if an upgrader is needed?
> 
> https://github.com/streamsets/datacollector/blob/master/kafka_target-protolib/src/main/java/com/streamsets/pipeline/stage/destination/kafka/KafkaTargetUpgrader.java

Thanks for the response, KafkaTargetUpgrader.java looks like it related to Kafka Producer. But changes are related to kafka_multisource.

SandishKumarHN:
> I also wonder if instead of making this an option, we just remove this entirely. At the pipeline level you can pick at least once or at most once.

Yeah, this makes sense. I will make changes.

kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
```diff
@@ -313,12 +316,17 @@ private Properties getKafkaProperties(Stage.Context context) {
...
     props.setProperty(KafkaConstants.KEY_DESERIALIZER_CLASS_CONFIG, conf.keyDeserializer.getKeyClass());
     props.setProperty(KafkaConstants.VALUE_DESERIALIZER_CLASS_CONFIG, conf.valueDeserializer.getValueClass());
     props.setProperty(KafkaConstants.CONFLUENT_SCHEMA_REGISTRY_URL_CONFIG, StringUtils.join(conf.dataFormatConfig.schemaRegistryUrls, ","));
 
+    if(conf.enableAutoCommitOffset) {
+      props.setProperty(KafkaConstants.AUTO_COMMIT_OFFEST, "true");
+      props.setProperty(KafkaConstants.AUTO_COMMIT_INTERVAL, String.valueOf(conf.autoCommitInterval));
+    } else {
+      props.setProperty(KafkaConstants.AUTO_COMMIT_OFFEST, "false");
+    }
```
jarcec: Similarly, as @brockn mentioned in the comments, I also believe we should not even expose this config and we should commit the offsets based on the pipeline's delivery guarantee configuration. The trouble is that it's the framework that knows when a change should be merged and for multi-threaded pipelines (this case) the framework doesn't support a call back to the origin "hey should commit now" as we have in the single-threaded pipelines. Updating the framework won't be easy though, so perhaps in the interim we could use the getContext.getDeliveryGuarantee() to figure out what is the configuration and either commit before/after calling getContext.processBatch() - what do you think @SandishKumarHN ?

SandishKumarHN: done

SandishKumarHN:
> Thanks for the submission @SandishKumarHN . I will need to also ask you to write an integration test for such a change - for the time being we keep the integration tests in a separate repository https://github.com/streamsets/datacollector-tests

kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
SandishKumarHN: for integration tests, I will work and create PR at source repository.

jarcec: Wonderful, thank you sir!

kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
```diff
@@ -313,12 +317,17 @@ private Properties getKafkaProperties(Stage.Context context) {
...
     props.setProperty(KafkaConstants.KEY_DESERIALIZER_CLASS_CONFIG, conf.keyDeserializer.getKeyClass());
     props.setProperty(KafkaConstants.VALUE_DESERIALIZER_CLASS_CONFIG, conf.valueDeserializer.getValueClass());
     props.setProperty(KafkaConstants.CONFLUENT_SCHEMA_REGISTRY_URL_CONFIG, StringUtils.join(conf.dataFormatConfig.schemaRegistryUrls, ","));
 
+    if (getContext().getDeliveryGuarantee() == DeliveryGuarantee.AT_MOST_ONCE) {
+      props.setProperty(KafkaConstants.AUTO_COMMIT_OFFEST, "true");
+      props.setProperty(KafkaConstants.AUTO_COMMIT_INTERVAL, "1000");
+    } else if (getContext().getDeliveryGuarantee() == DeliveryGuarantee.AT_LEAST_ONCE) {
+      props.setProperty(KafkaConstants.AUTO_COMMIT_OFFEST, "false");
+    }
```
jarcec: What if we would drop the auto commit in all cases and in do commit for AT_MOST_ONCE before we call the processBatch()? Or do you think that from performance perspective it's better to keep the auto.commit in such case?

SandishKumarHN: can we achieve AT_MOST_ONCE without changing Kafka properties?   are you suggesting making change for AT_MOST_ONCE at call() method?

jarcec: Yeah pretty much I'm thinking about handling the AT_MOST_ONCE here:

              if (list.size() == conf.maxBatchSize) {
                sendBatch(list);

And call the commit before sendBatch(list) What do you think?

jarcec:
Do you need any help with the integration tests or is all progressing well @SandishKumarHN ?

kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
SandishKumarHN: And keep value for AUTO_COMMIT_INTERVAL default value? which is 5000 ms

SandishKumarHN: if we commit before, would streamsets cache's data if pipeline crashs ? and avaoid duplicates ? 
`sendBatch(list)`

SandishKumarHN:
> Do you need any help with the integration tests or is all progressing well @SandishKumarHN ?

Sorry, didn't have much time to work on integration tests, planning to work finish this weekend.

kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/MultiKafkaSource.java
jarcec: So we have two modes of operation. "AT_LEAST_ONCE" and "AT_MOST_ONCE". Generally the framework does the following (not so well for multithreaded pipelines):

if(AT_MOST_ONCE) commit
runBatch()
if(AT_LEAST_ONCE) commit

Thus I would replicate the similar behaviour here in the origin and skip the autocommit completely. What do you think?

SandishKumarHN: AT_MOST_ONCE suppose to commit at a certain interval right? if we commit offset at every batch wouldn't that slow down the latency for smaller batches?

jarcec: AT_MOST_ONCE needs to guarantee no duplicates, so we have can't depend on the periodical commit as we could create duplicate this way.

SandishKumarHN: make sense, even we are looking for something like that.

jarcec:
How are things going @SandishKumarHN , anything I can do to help?

SandishKumarHN:
> How are things going @SandishKumarHN , anything I can do to help?

I have updated the PR with the latest changes and working on integration tests

jarcec:
Cool, let me know if there is anything that I can help with then.

jarcec:
Any luck @SandishKumarHN ? I'm happy to help.

SandishKumarHN:
> Any luck @SandishKumarHN ? I'm happy to help.

Sorry for late reply, was stuck with some issues. will have something this weekend.

jarcec:
I'm thinking that the proper test will be to have pipeline reading from Kafka, through delay processor terminating with scripting processor that throws exception depending on pipeline parameter. We can run the pipeline once, wait for it to fail, then second time, wait for it succeed and ensure that we read all the records from Kafka and nothing was lost. What do you think @SandishKumarHN ?

jarcec:
I took the liberty of jumping in and help writing the test case described above @SandishKumarHN . It actually uncovered a small issue in the patch here, so let me upload an updated variant.

Can I ask you to sign our NDA so that I can merge the fix?

SandishKumarHN:
@jarcec Sorry for making delay on test cases was quite busy with work. and I have already signed NDA and this is my second PR for datacollector project.

jarcec:
No worries, we all have day jobs @SandishKumarHN  - This patch has been merged, thank you for your contribution!

PR #89, opened 2019-03-28
SDC-11205 Fix pagination with relative URIs

asolfre:
- Adds a new configuration field to define the prefix to prepend to the next page url received on the response. Having a field instead of extracting the domain from the conf.resourceUrl makes it more flexible in case there is APIs where the relative path returned need more than the domain to be prepended.
- Add test case to cover the relative uri pagination

Issue link: https://issues.streamsets.com/browse/SDC-11205

### DIFF
--- a/basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/http/HttpClientSource.java
+++ b/basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/http/HttpClientSource.java
@@ -625,9 +625,9 @@ public class HttpClientSource extends BaseSource {
           RecordEL.setRecordInContext(stopVars, record);
           haveMorePages = !stopEval.eval(stopVars, conf.pagination.stopCondition, Boolean.class);
           if (haveMorePages) {
-        	final String nextPageURLPrefix = StringUtils.isNotBlank(conf.pagination.nextPageURLPrefix) ? conf.pagination.nextPageURLPrefix : "";
-        	final String nextPageFieldValue = record.get(conf.pagination.nextPageFieldPath).getValueAsString();	
-        	final String nextPageURL = nextPageFieldValue.startsWith(nextPageURLPrefix) ? nextPageFieldValue : nextPageURLPrefix.concat(nextPageFieldValue);
+            final String nextPageURLPrefix = StringUtils.isNotBlank(conf.pagination.nextPageURLPrefix) ? conf.pagination.nextPageURLPrefix : "";
+            final String nextPageFieldValue = record.get(conf.pagination.nextPageFieldPath).getValueAsString();
+            final String nextPageURL = nextPageFieldValue.startsWith(nextPageURLPrefix) ? nextPageFieldValue : nextPageURLPrefix.concat(nextPageFieldValue);
             next = Link.fromUri(nextPageURL).build();
           } else {
             next = null;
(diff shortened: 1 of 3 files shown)


basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/http/HttpClientSource.java
```diff
@@ -625,7 +625,10 @@ String parseResponse(long start, int maxRecords, BatchMaker batchMaker) throws S
           RecordEL.setRecordInContext(stopVars, record);
           haveMorePages = !stopEval.eval(stopVars, conf.pagination.stopCondition, Boolean.class);
           if (haveMorePages) {
-            next = Link.fromUri(record.get(conf.pagination.nextPageFieldPath).getValueAsString()).build();
+        	final String nextPageURLPrefix = StringUtils.isNotBlank(conf.pagination.nextPageURLPrefix) ? conf.pagination.nextPageURLPrefix : "";
+        	final String nextPageFieldValue = record.get(conf.pagination.nextPageFieldPath).getValueAsString();	
+        	final String nextPageURL = nextPageFieldValue.startsWith(nextPageURLPrefix) ? nextPageFieldValue : nextPageURLPrefix.concat(nextPageFieldValue);
```
jarcec: It seems like the pagination is a bit off.

basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/http/PaginationConfigBean.java
```diff
@@ -39,6 +39,18 @@
...
+      required = false,
+      type = ConfigDef.Type.STRING,
+      label = "Next Page Link Prefix",
+      description = "Prefix to concatenate with the value resolved from the Next Page Link Field. To be used in case the  Next Page Link is relative.",
+      group = "#0",
+      dependsOn = "mode",
+      triggeredByValue = "LINK_FIELD",
+      displayPosition = 30
+  )
+  public String nextPageURLPrefix;
```
jarcec: I wonder if we can get the next page URL from the primary URL and not force the user to specify it explicitly?

asolfre: The rationale behind it is to be more flexible with different relative pagination implementations.

Having a record URL like this one **https://www.somedomain.com/api/v1/foo/bar/status/?skip=0&metric=123&limit=100** ideally the relative next page URL will look like **/api/v1/foo/bar/status/?skip=100&metric=123&limit=100** and it's fairly easy to extract the domain from the primary URL and prepend it with the next page URL. But what happens if instead the API that it's being consumed returns the next page URL with the following formats:

- **v1/foo/bar/status/?skip=100&metric=123&limit=100**
- or **/status/?skip=100&metric=123&limit=100**
- or even **bar/status/?skip=100&metric=123&limit=100**

Letting users specify the prefix allows for them to be able to cover the different scenarios by entering the missing parts of the URL:
- **https://www.somedomain.com/api**
- **https://www.somedomain.com/api/v1/foo/bar**
- **https://www.somedomain.com/api/v1/foo/**

I haven't seen many example implementations of relative next page URLs and probably most of them will follow the first pattern (/api/v1/foo/bar/status/?skip=100&metric=123&limit=100) but this way the HTTP Stage wins flexibility in exchange of the user having to specify just an extra value.
Does it make sense?

basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/http/HttpClientSource.java
asolfre: Should be fixed now.

basic-lib/src/main/java/com/streamsets/pipeline/stage/origin/http/PaginationConfigBean.java
jarcec: Yes, it does, thanks for the explanation. It seems that the pagination is "everybody does their own thing" that at some it will be easier to instruct folks to use scripting processors to generate the "right (TM)" URL :-/

jarcec:
I've pushed the change through our internal gerrit to see if all tests properly pass.

jarcec:
I've merged the change, thank you for your contribution @asolfre !

PR #90, opened 2019-04-07
SDC-11243 Add "JDBC Driver Class Name" config to MySQL Binary Log origin

jsharper:
Useful for working around driver loading issues when using multiple JDBC origins (eg. when a pipeline using Oracle CDC Client origin starts before a pipeline using this origin.)

### DIFF
--- a/mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/Errors.java
+++ b/mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/Errors.java
@@ -26,7 +26,8 @@ public enum Errors implements ErrorCode {
   MYSQL_004("Error processing MySql event {} at offset {}: {}"),
   MYSQL_006("MySql server error: {}"),
   MYSQL_007("Ignore tables format error: {}"),
-  MYSQL_008("Include tables format error: {}"),;
+  MYSQL_008("Include tables format error: {}"),
+  MYSQL_009("Driver class not found: {}"),;
   private final String msg;
 
   Errors(String msg) {
--- a/mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
+++ b/mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
@@ -142,6 +142,16 @@ public abstract class MysqlSource extends BaseSource {
     hikariConfig.setPassword(getConfig().password);
     hikariConfig.setReadOnly(true);
     hikariConfig.addDataSourceProperty("useSSL", getConfig().useSsl);
+    if (getConfig().driverClassName != null && !getConfig().driverClassName.isEmpty()) {
+      try {
+        Class.forName(getConfig().driverClassName);
+      } catch (ClassNotFoundException e) {
+        LOG.error("Driver class not found: {}", e.getMessage(), e);
+        issues.add(getContext().createConfigIssue(
+            Groups.ADVANCED.name(), null, Errors.MYSQL_009, e.getMessage(), e
+        ));
+      }
+    }
     try {
       dataSource = new HikariDataSource(hikariConfig);
       offsetFactory = isGtidEnabled()
--- a/mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
+++ b/mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
@@ -147,6 +147,16 @@ public class MysqlSourceConfig {
   public boolean useSsl;
 
   @ConfigDef(
+      required = false,
+      type = ConfigDef.Type.STRING,
+      label = "JDBC Driver Class Name",
+      description = "Class name for pre-JDBC 4 compliant drivers.",
+      displayPosition = 85,
+      group = "ADVANCED"
+  )
+  public String driverClassName;
+
+  @ConfigDef(
       required = true,
       type = ConfigDef.Type.BOOLEAN,
       defaultValue = "false",


mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
```diff
@@ -142,6 +142,16 @@ public Record create(String recordSourceId) {
     hikariConfig.setPassword(getConfig().password);
     hikariConfig.setReadOnly(true);
     hikariConfig.addDataSourceProperty("useSSL", getConfig().useSsl);
+    if (getConfig().driverClassName != null && !getConfig().driverClassName.isEmpty()) {
```
jarcec: Is there guava on the classpath? Perhaps Strings.isNullOrEmpty()?

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
```diff
@@ -146,6 +146,16 @@
...
 
+  @ConfigDef(
+      required = false,
+      type = ConfigDef.Type.STRING,
+      label = "JDBC Driver Class Name",
+      description = "Class name for pre-JDBC 4 compliant drivers.",
+      displayPosition = 85,
+      group = "ADVANCED"
+  )
+  public String driverClassName;
```
jarcec: This is is for MySQL only, what about pre-loading it with MySQL value? It's unlikely that a user would need to change it right?

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/Errors.java
```diff
@@ -26,7 +26,8 @@
   MYSQL_004("Error processing MySql event {} at offset {}: {}"),
   MYSQL_006("MySql server error: {}"),
   MYSQL_007("Ignore tables format error: {}"),
-  MYSQL_008("Include tables format error: {}"),;
+  MYSQL_008("Include tables format error: {}"),
+  MYSQL_009("Driver class not found: {}"),;
```
jarcec: Let's put the semicolon on it's own line.

jarcec:
Have you by any chance seen my comments @jsharper ?

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/Errors.java
jsharper: done

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
jsharper: I was just following the conventions used elsewhere in this code.  Unfortunately, I don't have time to dig into whether guava is in classpath for this module or do another test build to experiment or validate.  I found no other current use of guava in this directory.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
jsharper: The mysql jdbc drivers have used different class name values over time.  Given that, I'm not sure it is appropriate to assume the value that is correct for the driver that the end-user will supply.

If you want to go down that path, I think it would be great if a mysql jdbc driver was actually just included in the datacollector build and distribution (I believe it's even available in the maven repository, so this may not be difficult).  In that case, the class name could just be hardcoded and the whole issue could be addressed with a single Class.forName line; this config parameter would not be needed at all.  Even better, end users would not have to install a mysql driver themselves.

I'm not sure if there are license compatibility issues with doing this, though.  I'll leave that up to you guys to consider.

jarcec: We sadly can't include the MySQL driver exactly for the reason you're describing - the license doesn't allow us to distribute the jar file ourselves.

You're right that the driver can potentially change the name and it did some time ago, but it's constant for several past years, so I don't really mind prepopulating the current value.

jsharper: I added a default value of "com.mysql.cj.jdbc.Driver"

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
jarcec: Fair enough, I'm fine with consistency.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
```diff
@@ -142,6 +142,16 @@ public Record create(String recordSourceId) {
...
+      try {
+        Class.forName(getConfig().driverClassName);
+      } catch (ClassNotFoundException e) {
+        LOG.error("Driver class not found: {}", e.getMessage(), e);
+        issues.add(getContext().createConfigIssue(
+            Groups.ADVANCED.name(), null, Errors.MYSQL_009, e.getMessage(), e
```
jarcec: My apologies, I just noticed, the null shouldn't be null but should be "config.driverClassName", right?

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
```diff
@@ -146,6 +146,17 @@
   )
   public boolean useSsl;
 
+  @ConfigDef(
+      required = false,
+      type = ConfigDef.Type.STRING,
+      defaultValue = "com.mysql.cj.jdbc.Driver",
```
jarcec: Isn't the default driver name com.mysql.jdbc.Driver ?

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSource.java
jsharper: Fixed.

mysql-binlog-lib/src/main/java/com/streamsets/pipeline/stage/origin/mysql/MysqlSourceConfig.java
jsharper: Not for current versions of Connector/J:

https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-driver-name.html
https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-usagenotes-connect-drivermanager.html

jsharper:
Are there any more concerns with this PR or this approach to fixing SDC-11243?

PR #91, opened 2019-04-26
Sdc 11356

zlosim:
Storing schema ID to record attribute when using registry for avro schemas


jarcec:
The change looks good to me in general @zlosim ! Can I ask you to write an STF test for it too to ensure that it's working properly and we won't regress in the future?

zlosim:
Hi @jarcec 
sure i will try, but I`m not quite sure what an STF test is. If you can point me in right direction I will start ASAP, thanks!

jarcec:
Oh absolutely! We're building set of integration tests that we are calling STF @zlosim . For the time being they are living in separate repository here:

https://github.com/streamsets/datacollector-tests

How to start guides are here:

https://streamsets.com/blog/introducing-the-streamsets-test-framework/
https://streamsets.com/resources/video/a-short-course-on-the-streamsets-test-framework/

Running the tests currently requires a license which I will send your way later today.

jarcec:
I've just forwarded you the license @zlosim .

jarcec:
Any update on the integration tests @zlosim ?

zlosim:
Sorry @jarcec for the late reply
I replied to your mail because the link with the license was already expired :( If you could resend it I`ll try to wrap it up during the weekend

zlosim:
Hi @jarcec I think I`m ready with the integration test and would like to test it somehow before pull request if you can send my the license one more time. Thanks

PR #93, opened 2019-05-03
Using JSON Entities breaks scroll functionality in Elasticsearch 1.x

ryantargz:
Using query params instead, and is supported past 1.x

You can find an example of how 1.x supports scroll params here:
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/search-request-scroll.html

scroll_id is supported as either a raw text entity (no JSON formatting) or a query param.

Note the following quote:
"The returned _scroll_id attribute can be passed in the request body or in the query string as ?scroll_id=...."

Here is the 5.2 doc: 
https://www.elastic.co/guide/en/elasticsearch/reference/5.2/search-request-scroll.html

Which states that query params still work in future versions past 1.x in lieu of a JSON entity:
"The scroll_id can also be passed as a query string parameter or in the request body."

### DIFF
--- a/elasticsearch-protolib/src/main/java/com/streamsets/pipeline/lib/elasticsearch/ElasticsearchStageDelegate.java
+++ b/elasticsearch-protolib/src/main/java/com/streamsets/pipeline/lib/elasticsearch/ElasticsearchStageDelegate.java
@@ -197,6 +197,15 @@ public class ElasticsearchStageDelegate {
     return restClient.performRequest(method, endpoint, params, entity, headers);
   }
 
+  public Response performRequest(
+          String method,
+          String endpoint,
+          Map<String, String> params,
+          Header... headers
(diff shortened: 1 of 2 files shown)


jarcec:
We do not support Elasticsearch 1.x since it's a really old version. I worry about opening the door for adding the support as that might mean on us to keep that compatibility which will be an additional burden. Thus I apologize, but I will have to reject the change for now.

ryantargz:
@jarcec 
I was under the impression that 1.x was supported. See doc: 

https://streamsets.com/documentation/datacollector/latest/help/datacollector/UserGuide/Installation/AddtionalStageLibs.html#concept_evs_xkm_s5

> For Elasticsearch 1.x, 2.x, and 5.x.

If that is not the case, 1.x should be removed to avoid confusion by me and others down the road.

PR #95, opened 2019-05-13
Added ability to override the header row or provide one if it is miss…

erichardson-icct:
…ing.

Commons CSV allows  A CSVFormat to override the header row or to provide the headers if there is no header row. see https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#withHeader-java.lang.String...-

This patch will allow a Streamsets DataFormat to use that capability.


erichardson-icct:
Recolves [SDC-8637](https://issues.streamsets.com/browse/SDC-8637)

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
```diff
@@ -275,6 +275,20 @@
...
+      label = "Header Line Override",
+      description = "The String to use as the header line, must have same delimiter as specified above.",
+      displayPosition = 384,
+      group = "DATA_FORMAT",
+      dependencies = {
+          @Dependency(configName = "dataFormat^", triggeredByValues = "DELIMITED"),
+          @Dependency(configName = "csvHeader", triggeredByValues = {"IGNORE_HEADER","NO_HEADER"})
+      }
+  )
+  public String csvOverrideHeader;
```
jarcec: I wonder, wouldn't it be better to declare this as List<String> rather than a CSV fragment?

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
```diff
@@ -275,6 +275,20 @@
   @ValueChooserModel(CsvHeaderChooserValues.class)
   public CsvHeader csvHeader = CsvHeader.NO_HEADER;
```
jarcec: I wonder if making a new CsvHeader.EXPLICIT_HEADER would make sense and then display (and use) csvOverrideHeader only in that case?

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
erichardson-icct: My thought when doing it with the delimited fragment was to enable cut and paste from a source. Most useful when you want to split the input into lines, transform the lines (ex regex replacement to correct data errors) then Parse Data Processor get the data.
I could modify it to allow either list or delimited fragment.

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
erichardson-icct: Slight problem. CsvHeader enum has two pieces of information in it three values. 
```
WITH_HEADER("With Header Line"),
IGNORE_HEADER("Ignore Header Line"),
NO_HEADER("No Header Line"),
```
1. The file has a header line (or not)
1. The header line should be used (or not)

The use explicit header line has to work as expected when there is no header line and when the header line should be ignored. This would lead to two new values like
```
NO_HEADER_USE_EXPLICIT("No Header Line, use configured explicit header"),
IGNORE_HEADER_USE_OVERRIDE("Ignore Header Line, use configured override header"),
```

My opinion was that leaving the current 3 states as-is (because two of them accurately represent the possible states as used) and show a configuration error when directing the system to both use the header and use the explicit header was the least confusing.

It also fits more closely to the current situation, which is an implicit header  1, 2, 3 etc. Leave it at 3 values and NO_HEADER and IGNORE_HEADER can use either the implicit header or the override header

jarcec: I see what you're saying. When I was initially reading the text, I was thinking that specifying header is applicable only for "NO_HEADER" and hence the "EXPLICIT_HEADER" would be a special variant of NO_HEADER. 

I see the reasoning to allow specify custom header with IGNORE_HEADER though, so let's keep it this way.

commonlib/src/main/java/com/streamsets/pipeline/stage/origin/lib/DataParserFormatConfig.java
jarcec: If you don't mind, I would prefer to have a list here - it will be easier to operate and will also better fit our framework.

erichardson-icct: Just pushed the commit.

metadaddy:
@eric-richardson-g2o Are you still interesting in pushing this forward? It looks like you just need to sign the CLA.

eric-richardson-g2o:
I signed the CLA. Still not passing the CLA check, but when I follow the link to the CLA to sign it again, it registers as already signed.

eric-richardson-g2o:
OK, post a comment and the CLA check suddenly passes...

PR #99, opened 2019-07-22
Update RabbitExchangeConfigBean.java

klaxman:
Added elDef to allow record functions in routing key attribute. Please review.


klaxman:
Allow record functions

PR #103, opened 2019-10-09
[SDC-12625] Update the default maximum number of worker threads for Kudu stages

OlegPt:


dimaspivak:
This default would no longer be consistent with the help text and can cause upgrade compatibility issues. Since users can set the value on their own, I don't think this is a worthwhile change to make at this time.

PR #106, opened 2019-11-11
Update azure-lib to allow domain name to be selected.

tgillot:
Currently StreamSets does not allow the Azure Environment domain name to be selected, which means it will not work with non-public national clouds (e.g. usgovcloudapi.net). By default the public cloud domain name is used.

The azure-lib was updated to use the setEndpoint method and a new definition was added to allow the domain name to be entered in the UI.

 See the following Microsoft documentation for more info:
https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.eventhubs.connectionstringbuilder.setendpoint?view=azure-java-stable

### DIFF
--- a/azure-lib/src/main/java/com/streamsets/pipeline/lib/eventhubs/EventHubConfigBean.java
+++ b/azure-lib/src/main/java/com/streamsets/pipeline/lib/eventhubs/EventHubConfigBean.java
@@ -44,6 +44,17 @@ public class EventHubConfigBean {
   @ConfigDef(
       required = true,
       type = ConfigDef.Type.STRING,
+      label = "Environment Domain Name",
+      defaultValue = "",
+      description = "Domain name of the Azure environment",
+      displayPosition = 20,
+      group = "EVENT_HUB"
+  )
+  public String domainName = "";
+
+  @ConfigDef(
+      required = true,
+      type = ConfigDef.Type.STRING,
       label = "Shared Access Policy Name",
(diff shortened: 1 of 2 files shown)


tgillot:
Also, this is a fix for this Jira bug:
https://issues.streamsets.com/browse/SDC-12134

azure-lib/src/main/java/com/streamsets/pipeline/lib/eventhubs/EventHubConfigBean.java
```diff
@@ -42,6 +42,17 @@
...
+  @ConfigDef(
+      required = true,
+      type = ConfigDef.Type.STRING,
+      label = "Environment Domain Name",
+      defaultValue = "",
+      description = "Domain name of the Azure environment",
+      displayPosition = 20,
+      group = "EVENT_HUB"
+  )
+  public String domainName = "";
```
jarcec: Since this config won’t be applicable for most users, do you mind adding a checkbox “Custom Domain Name” (or something like that) and display this configuration only if that checkbox is set? And then similarly apply the setEndpoint only if the checkbox is configured.

metadaddy: @tgillot I'm trying to figure out how we could test this without access to one of the non-public clouds. Will the same mechanism work with the regular Azure endpoint?

PR #120, opened 2020-01-24
SDC-13312. Handle empty string as null

ystory:
The purpose of this pull request is to avoid finding an empty string key in S3 bucket.


metadaddy:
Merged as 383ae3b. @jarcec could you close?

ystory:
Thanks. @metadaddy :)

PR #122, opened 2020-02-14
SDC-12512 Bug Fix

cuijunhengdtc:
Pipelines (which were running before restart) are stuck in state `Starting`.
The root cause is that the snapshot name is always "default", so it will cause an Exception, and the state will not change.
I change the snapshot name, and add the uuid to the name. 
So it will work correctly. And the pipeline state will change to "Running".


cuijunhengdtc:
pipeline_restart_snapshot_error
The error trace before the bug fix.
And after the fix, it can work normally.

cuijunhengdtc:
restart_snapshot
After the fix, the snapshot list.

ayushmandey97:
Can this be merged? The logic looks good, and this issue is causing issues in our production data collector, upon restarts.

metadaddy:
Hi @ayushmandey97 - I've asked engineering to take a look. In the meantime, if you're a StreamSets customer, please open a support ticket referencing SDC-12512 and this pull request.

metadaddy:
Merged as 383ae3bbbb8985043e3725d944f24eec3dea4fb3.  @jarcec could you close?

PR #131, opened 2020-11-01
SDC-16127: Fix kafka-multi NullPointer on null payload

chapsuk:
Hi all! Thanks for the awesome product)

### Description:

When we get kafka message like this
```
key: someKey value: null
```
Kafka Multitopic origin throws an Exception
```
ERROR MultiKafkaSource - Encountered error in multi kafka thread 0 during read null
java.lang.NullPointerException
	at com.streamsets.pipeline.lib.util.AvroSchemaHelper.detectSchemaId(AvroSchemaHelper.java:238)
```

### Fix

Handle `null payload` with stage error handler


kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/KafkaMultitopicRunnable.java
```diff
@@ -215,6 +215,19 @@ private void commitSyncAndProcess() {
     String messageId = getMessageId(topic, partition, offset);
     List<Record> records = new ArrayList<>();
 
+    if (payload == null) {
+      Record record = getContext().createRecord(messageId);
+      record.set(Field.create(item.value()));
+      errorRecordHandler.onError(
```
xbcratos, line 221: Set the correct formatting for this line. It should be (all in a single line as it fits in the length of one line): 
errorRecordHandler.onError(new OnRecordErrorException(record, KafkaErrors.KAFKA_74, messageId));

xbcratos (changes requested):

kafka_multisource-protolib/src/main/java/com/streamsets/pipeline/stage/origin/multikafka/KafkaMultitopicRunnable.java
chapsuk: done

xbcratos:
After some analysis we have determined null payloads are okay. See: [kafka tombstones](https://medium.com/@damienthomlutz/deleting-records-in-kafka-aka-tombstones-651114655a16)

So we have modified the patch and merged it in master branch. See: https://github.com/streamsets/datacollector/commit/3092137d0db55a4c392b1d138e2595e999ae788c

Hence closing this pull request.

Thanks!

chapsuk:
Thanks, but keep in mind that single topic consumer [handle null payload as error](https://github.com/streamsets/datacollector/blob/master/kafka_source-protolib/src/main/java/com/streamsets/pipeline/stage/origin/kafka/BaseKafkaSource.java#L284-L295), this difference may be confuse users.

PR #136, opened 2020-11-16
SDC-16191: Allow BYTE_ARRAY fields to be written to BLOB (Oracle) database columns

mss:
Ironically, `BLOB` columns are not recognized as binary columns and thus writing a `BYTE_ARRAY` to such a column fails.  At least for Oracle where the fallback `setObject()` call doesn't work but `setBytes()` is totally fine.

This change makes the JDBC Producer and friends recognize BLOB columns as binary fields and also makes the mapping fields in the object `static final`.


jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
```diff
@@ -69,13 +69,15 @@
...
+      Types.BLOB,
       Types.LONGVARBINARY,
       Types.VARBINARY
-  ));
-  private Set<Integer> textTypes = new HashSet<>(Arrays.asList(
+  )));
+  private static final Set<Integer> TEXT_TYPES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
       Types.CHAR,
       Types.VARCHAR,
       Types.BLOB,
```
jarcec, line 83: Shouldn't we remove the BLOB from this type?

jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
```diff
@@ -69,13 +69,15 @@
   private List<String> primaryKeyColumns;
   private Map<String, String> columnsWithoutPrimaryKeys;
   JdbcRecordReader recordReader;
+
   /* Static lists for column type checking in setParamsToStatement */
-  private Set<Integer> binaryTypes = new HashSet<>(Arrays.asList(
+  private static final Set<Integer> BINARY_TYPES = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
```
jarcec, line 74: Can I ask you to not do the unrelated changes, please? I understand the logic in general, but it makes it difficult for us to trace git-blame and cherry-pick when patches have changes that are not necessary for them.

jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
mss: I am not sure, in the end I decided to keep this in here. The logic of these sets is that when the software sees a record field of the type as given by the set's name it checks if the mapped database column is in there. If this is the case it uses the type specific `setFoo` JDBC method, else `setObject`.

In this case when it sees a STRING field and a BLOB database column it would try to use `setString` instead of `setObject`. This doesn't work for Oracle databases but there might be other databases/JDBC drivers where this might work.

I didn't want to change that behaviour, just add the logic to use `setBytes` instead of setObject when it sees a BYTE_ARRAY field.

jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
mss: Sure, that's why I put this into a separate commit so the other change is easy to cherry-pick and remove from this PR if it isn't wanted. Do you want me to remove it from the PR?

jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
jarcec: Makes sense, thanks for the explanation.

jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
jarcec: Yes please, do remove that from the PR so that we can take the PR as a whole and not cherry-pick individual commits.

jarcec:
Hmm... you should have gotten an update from CLAassistant to sign an Individual Contributor License Agreement @mss  - usually, I can see comments on the PR which is not the case here. Can you validate for me if you received the email? If no I will have to figure out why the bot is not jumping in right now.

jdbc-protolib/src/main/java/com/streamsets/pipeline/lib/jdbc/JdbcBaseRecordWriter.java
mss: Done, I rebased on master and dropped the commit.

mss:
> Hmm... you should have gotten an update from CLAassistant to sign an Individual Contributor License Agreement @mss - usually, I can see comments on the PR which is not the case here. Can you validate for me if you received the email? If no I will have to figure out why the bot is not jumping in right now.

The CLA one can be found on my other PR #134 :-)

jarcec:
[SDC-16191](https://issues.streamsets.com/browse/SDC-16191) was merged, thank you for your contribution @mss !

